PR #1, opened 2021-11-23
Review 1

bbenligiray:
A fake PR to comment on things


index.ts
```diff
@@ -0,0 +1,217 @@
...
+import * as path from "path";
+import { safeDecode } from "./node/abi-encoding";
+//TODO: remove and use @api3/airnode-node import
+import { buildEVMProvider } from "./node/evm-provider";
+//TODO: remove and use @api3/airnode-node import
+import { removeKey, removeKeys } from "./node/object-utils";
+//TODO: remove and use @api3/airnode-node import
+import { getReservedParameters, RESERVED_PARAMETERS } from "./node/parameters";
+//TODO: remove and use @api3/airnode-node import
+import { deriveSponsorWallet } from "./node/wallet";
```
bbenligiray, line 17: Are these even exported?

index.ts
```diff
@@ -0,0 +1,217 @@
...
+//TODO: remove and use "@api3/airnode-protocol" import;
+import RrpBeaconServer from "./RrpBeaconServer.json";
```
bbenligiray, line 19: I guess you would import the contract and build it yourself

index.ts
```diff
@@ -0,0 +1,217 @@
...
+        const airnodeRrp = AirnodeRrpFactory.connect(
+          chain.contracts.AirnodeRrp,
+          provider
+        );
+
+        // TODO: use factory class to create contract instead
+        //   const rrpBeaconServer = RrpBeaconServerFactory.connect(RrpBeaconServer.address, provider);
+        const abi = RrpBeaconServer.abi;
+        const rrpBeaconServer = new ethers.Contract(
+          chain.contracts.RrpBeaconServer,
```
bbenligiray, line 56: That's cool, I was wondering where we would put this

index.ts
bbenligiray: I see that you're considering exporting the factory, that would also work (but it's a hassle to export factories for all the contracts we develop so we have to draw the line somewhere)

index.ts
```diff
@@ -0,0 +1,217 @@
...
+          abi,
+          provider
+        );
+
+        const airnodeWallet = ethers.Wallet.fromMnemonic(
+          nodeSettings.airnodeWalletMnemonic
+        ).connect(provider);
+
+        // TODO: should templateId come from triggers or somewhere else?
```
bbenligiray, line 65: We can have a `triggers.keeperJobs`

index.ts
```diff
@@ -0,0 +1,217 @@
...
+            } catch (e) {
+              console.log("Error: failed to extract data from API response");
+              return;
+            }
+
+            // **************************************************************************
+            // 4. Read beacon
+            // **************************************************************************
+            const beaconResponse = await rrpBeaconServer
+              .connect(airnodeWallet)
```
bbenligiray, line 132: This will probably be `VoidSigner` with `address(0)`

index.ts
```diff
@@ -0,0 +1,217 @@
...
+              return;
+            }
+
+            const deviation = delta
+              .mul(100 * Number(reservedParameters._times)) // TODO: can _times be null or 0?
+              .div(apiValue);
+            console.log(
+              "Info: deviation %",
+              deviation.toNumber() / Number(reservedParameters._times)
+            );
```
bbenligiray, line 157: I wonder if we should do this with a static call to https://github.com/api3dao/airnode/blob/psp-draft-v4/packages/airnode-protocol/contracts/psp/BeaconServer.sol#L326
Probably not though

index.ts
```diff
@@ -0,0 +1,217 @@
...
+            // 5. Check deviation
+            // **************************************************************************
+            const delta = beaconResponse.value.sub(apiValue).abs();
+            if (delta.eq(0)) {
+              console.log("Info: beacon is up-to-date. skipping update");
+              return;
+            }
+
+            const deviation = delta
+              .mul(100 * Number(reservedParameters._times)) // TODO: can _times be null or 0?
```
bbenligiray, line 152: Yes

index.ts
bbenligiray: I guess `triggers.rrpBeaconServerKeeperJobs` would be more accurate

index.ts
```diff
@@ -0,0 +1,217 @@
...
+              deviation.toNumber() / Number(reservedParameters._times)
+            );
+
+            // **************************************************************************
+            // 6. Update beacon if necessary (call makeRequest)
+            // **************************************************************************
+
+            // TODO: should we calculate the requestId hash or find a way to prevent sending
+            // the same request more that once? RrpBeaconServer.requestIdToTemplateId keeps
+            // track of the pending requests using a templateId
```
bbenligiray, line 165: Good point, but the request ID we derive here will be different than the one we just made one minute ago because of the incremented nonce.

Let's talk about this at the call.

index.ts
```diff
@@ -0,0 +1,217 @@
...
+
+            // TODO: 5% is hardcoded, should this be read from config?
```
bbenligiray, line 167: Probably from `triggers.rrpBeaconServerKeeperJobs`

index.ts
```diff
@@ -0,0 +1,217 @@
...
+            const tolerance = 5;
+            if (deviation.lte(tolerance * Number(reservedParameters._times))) {
```
bbenligiray, line 169: It's not nice that these operations aren't purely BigNumber

index.ts
```diff
@@ -0,0 +1,217 @@
...
+             */
+            // console.log(
+            //   "🚀 ~ file: index.ts ~ line 161 ~ handler ~ await rrpBeaconServer.sponsorToUpdateRequesterToPermissionStatus()",
+            //   await rrpBeaconServer.sponsorToUpdateRequesterToPermissionStatus(
+            //     airnodeWallet.address,
+            //     airnodeWallet.address
+            //   )
+            // );
+
+            // TODO: who should be the sponsor?
```
bbenligiray, line 190: I suppose this is another config file thing, `sponsorWalletPath` set per job

bbenligiray: So we would enter `m/123/0`, `m/123/1`, `m/123/2` etc. for different jobs (or the same path for all the jobs)

andreogle:
I took a brief look at this repo and it looks like the compiled JS has been committed so there are duplicate files. You probably want to set `"ourDir": "./dist"` in tsconfig.json and add `dist/` to .gitignore

index.ts
```diff
@@ -0,0 +1,217 @@
...
+              ethers.utils.HDNode.fromMnemonic(
+                nodeSettings.airnodeWalletMnemonic
+              ),
+              airnodeWallet.address
+            ).address;
+
+            // TODO: why can't we send encoded parameters to be forwarded to AirnodeRrp?
+            // When using config.json.example we must pass a "from" parameter and the only
+            // way to get this request to work is if we add it a fixedParameter in the node
+            // config file
```
bbenligiray, line 202: The idea is that all parameters for a beacon is contained in the template, so the template has to be "full". Another thing to discuss at the call.

serverless.yml
```diff
@@ -0,0 +1,11 @@
+service: airkeeper
+
+provider:
```
bbenligiray, line 3: We'll probably need a decent memory limit

index.ts
bbenligiray: Overall, this is the only major issue

index.ts
acenolaza: They are not and that's why I temporarily copied the code over to airkeeper repo. I made the assumption that they are going to be exported in the airnode repo tho.

index.ts
acenolaza: Right, I was thinking about exporting the factory like with do for AirnodeRrp and a few others but I get your point.

index.ts
acenolaza: Yeah, I was thinking the same and I certainly don't trust JS with numbers and math operations. I'll try to cast everything to BigNumber and use that interface instead.

index.ts
acenolaza: 👍🏻 I left the comment there as reminder that we need to discuss [this](https://api3workspace.slack.com/archives/C027Y2FAV0S/p1637335207072900) comment you made.

serverless.yml
acenolaza: ✅

acenolaza: I'm also trying to add scheduling on a fixed rate to this file.

index.ts
acenolaza: That's actually an interesting idea. I forgot about that BeaconServer and it's architecture. We would need to create a subscription per keeper trigger I'm thinking 🤔

index.ts
bbenligiray: Okay, I guess that will be in 0.3.0?

index.ts
bbenligiray: I think this was about TypeChain

index.ts
acenolaza: I'm thinking that this config change should be done in airnode package with a validator update since Beacon Server  probably will not only be used by Airkeeper.

index.ts
acenolaza: I'm a bit confused about this `sponsorWalletPath` configuration. If the idea is that we `rrpBeaconServer.requestBeaconUpdate()` more than once with different sponsor wallets shouldn't we use a a sponsor address array instead per job?

Something like:

```
    "rrpBeaconServerKeeperJobs": [
      {
        "templateId": "0x50c604914d8ed35473149457a1a0912b785813b4e2e51bd2b75409ca25c50e1d",
        "oisTitle": "Currency Converter API",
        "endpointName": "convertToUSD",
        "deviationPercentage": "5",
        "sponsors": ["0x1234..., 0x5678, etc"]
      }
    ]
```

And Airkeeper would then derive the sponsor wallet for each sponsor in the `rrpBeaconServerKeeperJobs.sponsors` array and `call rrpBeaconServer.requestBeaconUpdate()`

index.ts
bbenligiray: If we're going to use a separate config file for airkeeper that complements config.json (which I think we had decided at the call yesterday) then we can put this field in that field instead of the actual config.json

index.ts
bbenligiray: The idea is that if I'm a sponsor, I don't want my (RRP) sponsor wallet to be used to request beacon updates because that wallet is already being used to fulfill RRP requests. So the idea here is that I specify the path of the wallet that I want the request updates to be made with.
btw don't confuse this with the sponsor/sponsorWallet parameters that the keeper should use while creating the request

index.ts
acenolaza: Yes, working on this today

bbenligiray:
@acenolaza Let me know when you want me to comment and I can open another PR

acenolaza:
> @acenolaza Let me know when you want me to comment and I can open another PR

Please do. I've covered all of the issues (except removing the files copied over from node until we publish the updated packages)

serverless.yml
```diff
@@ -0,0 +1,26 @@
...
+variablesResolutionMode: 20210219
+
+provider:
+  name: aws
+  # region: ${opt:region}
+  # stage: ${opt:stage}
+  runtime: nodejs12.x
+  memorySize: 512
+  lambdaHashingVersion: 20201221
+  logRetentionInDays: 14
```
Ashar2shahid, line 12: I guess can have longer logs but I guess that depends on users

src/index.ts
```diff
@@ -0,0 +1,323 @@
...
+            console.log("[ERROR]\tfailed to extract data from API response");
+            return;
+          }
+
+          // **************************************************************************
+          // 6. Read beacon
+          // **************************************************************************
+          console.log("[DEBUG]\treading beacon value from server...");
+          // address(0) is considered whitelisted
+          const voidSigner = new ethers.VoidSigner(
```
Ashar2shahid, line 159: this was pretty cool implementation, didn't know about this

config/airkeeper.json.example
```diff
@@ -0,0 +1,21 @@
+{
+  "chains": [
```
Ashar2shahid, line 2: I assume the chains are matched to the respective chain in airnode.json via the same index? I think it would be a good idea to specify the chainId here as well

Ashar2shahid (commented):
Some comments, Index.ts looks fine will test it with a beacon server and get back if I find any issues/ usability concerns

README.md
```diff
@@ -1,2 +1,50 @@
-# airkeeper
-A tool to update a beacon server value on a time interval
+# Airkeeper
+
+> A tool to update a beacon server value on a time interval
+
+This project is basically a makeshift version of the PSP protocol and it will be used to trigger beacon values updates on a fixed time interval of 1 minute.
```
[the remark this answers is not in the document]
acenolaza: I'm open to suggestions 😄

serverless.yml
```diff
@@ -0,0 +1,26 @@
+service: airkeeper
+
+variablesResolutionMode: 20210219
```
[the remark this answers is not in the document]
acenolaza: I'm not using SSM but I comment it out then I get this warning message:

```
Serverless: Deprecation warning: Variables resolver reports following resolution errors:
              - Cannot resolve variable at "provider.environment": Cannot parse "load-secrets-env.js": Approached a JS function resolver, confirm it's updated to work with a new parser by setting "variablesResolutionMode: 20210326" in service config. Falling back to old resolver
```

src/index.ts
acenolaza: Yes indeed. All credit goes to Burak for finding this: https://api3workspace.slack.com/archives/C027Y2FAV0S/p1637666522074300

config/airkeeper.json.example
acenolaza: This is a really good point. I'm not sure how merge does the matching but I'll test it with multiple chains in the config.

acenolaza: Indeed it's indexed based. I would really like to have the RrpBeaconServer contract address in the Airnode config.json and get rid of this config in Airkeeper. This would make it a lot less error prone in my opinion.

PR #4, opened 2021-12-16
BEC-41: Implement same gas price scheme from Airnode

acenolaza:
Ended up just copying over the gas price code to Airkeeper since those changes haven't yet been published in Airnode.


aquarat (approved):
LGTM 🥇
You essentially cherry-picked everything, very flattering :)

PR #5, opened 2021-12-20
BEC-40: Fetch transaction count for each keeperSponsorWallet

acenolaza:
1. Grouped rrpBeaconServerKeeperJobs by keeperSponsor address in order to run each job sequentially and be able to manually assign an incremental nonce for each one. These groups are run in parallel similar to [Airnode](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/evm/fulfillments/index.ts#L144).
2. I've also set the current block in the event fetching filters and tried to minimize roundtrips to external sources as much as possible.


src/start.ts
```diff
@@ -54,265 +54,331 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
-              JSON.stringify(apiResponse.data)
-            );
-            let message;
-            if (error instanceof Error) message = error.message;
-            else message = String(error);
-            console.log("[DEBUG]\tmessage:", message);
-            return;
-          }
+            // Fetch keeperSponsorWallet transaction count
+            let keeperSponsorWalletTransactionCount =
```
aquarat, line 84: I probably don't have all the context and all of this is from a place of trying to help:

`let keeperSponsorWalletTransactionCount` is updated once further down (`const nonce = keeperSponsorWalletTransactionCount++;`); if you converted the for loop to a `.map` and exposed the index argument you'd be able to say `const nonce = keeperSponsorWalletTransactionCount+sponsorIndex`. If the function handed to `.map` was `async` you'd get coroutine execution, which could be very useful. This should be fine because the wallets are separate. Airnode does this and it's probably why transactions were grouped like this.

So instead of 
```
            for (const {
              templateId,
              parameters,
              oisTitle,
              endpointName,
              deviationPercentage,
              requestSponsor,
              eventLogMaxBlocks,
            } of rrpBeaconServerKeeperJobs) {
```
it could be 
```
            rrpBeaconServerKeeperJobs.map(async ({
              templateId,
              parameters,
              oisTitle,
              endpointName,
              deviationPercentage,
              requestSponsor,
              eventLogMaxBlocks}, index) => {...});
```

src/start.ts
```diff
@@ -54,265 +54,331 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+              // **************************************************************************
+              // 4. Fetch template by ID
+              // **************************************************************************
+              console.log("[DEBUG]\tfetching template...");
+              let template: [string, string, string] & {
+                airnode: string;
+                endpointId: string;
+                parameters: string;
+              };
+              try {
```
aquarat, line 111: Airnode uses a small recursivey library called [`go`](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/utils/promise-utils.ts), which can handle retries and errors. It is named after Go's `go` action, but ironically bears little resemblance hah.

It is useful because it returns the result and error log as a tuple - so you don't need a try-catch.

This allows for logs to be returned by the function ([which Airnode also does](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/evm/gas-prices.ts)). The return type may look like `Promise<Logs[]>`.

Logging in cloud functions slows them down, so by deferring logging you'll speed things up potentially - the downside to this is that if the function times out the logs will be lost.

None of this is critical.

src/start.ts
```diff
@@ -54,265 +54,331 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+                if (requestIsAwaitingFulfillment) {
+                  console.log(
+                    "[INFO]\trequest is awaiting fulfillment. skipping update"
+                  );
+                  continue;
+                }
+              }
+
+              // Fetch current gas fee data
+              const [gasPriceLogs, gasTarget] = await getGasPrice({
```
aquarat, line 341: In this scenario you'll retrieve the gas pricing for every request, this will slow things down a lot and also increase the requests to the chain gateway/EVM node.

I suggest you move this up so the result can be re-used. This would be on a per-chain-provider basis, so probably to line ~46.

aquarat (commented):
As per comments (which have URLs and detail), you may wish to:
- look into using Airnode's `go` error and retry library
- convert the for loop at line 93 to a `map(async ({...}, index) => {...})`
- make the transaction count `const`
- make the nonce `const` and `= transactionCount + perSponsorIndex`
- move the gas retrieval to line ~92 such that it is cached per chain-provider

Just generally this is very readable and I enjoy your work.

src/start.ts
acenolaza: I would really like to revert `keeperSponsorWalletTransactionCount` to a being a `const` instead of a `let` but each rrpBeaconServerKeeperJob must be run sequentially since these will translate into 1 or more transaction per keeperSponsorWallet (similar to [here](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/evm/fulfillments/index.ts#L120) in node package).

I might be misunderstanding the suggested improvement tho 😄

src/start.ts
acenolaza: Good suggestion and I even have already thought about adding that library to Airkeeper. The thing is that I wanted to keep the code as simple as possible until the main requirements were met and the iterate over it to make it better. If you agree I'm going to create a ticket in BEC board to keep track of this improvement.

src/start.ts
acenolaza: This a good point but I had a discussion with Burak where he suggested that we only do this kind of chain requests whenever we are absolutely sure that we must create a beacon request transaction since most of the times Airkeeper will probably not reach this far down. Meaning that Airkeeper jobs running every minute will probably won't trigger a beacon update therefore no gas price data will be needed. Thoughts?

src/start.ts
aquarat: Sounds good :)

src/start.ts
aquarat: Ah, that's an interesting point.

The solution then would probably be to leave as-is for now but ultimately separate deviation assessment and possible result submission (sounds familiar) - and then cache gas pricing during the submission phase.

src/start.ts
aquarat: Haha, that makes sense, thanks for bearing with me :)

aquarat (approved):
All queries resolved :)

src/start.ts
```diff
@@ -5,7 +5,7 @@ import * as node from "@api3/airnode-node";
 import * as adapter from "@api3/airnode-adapter";
 import * as protocol from "@api3/airnode-protocol";
 import * as abi from "@api3/airnode-abi";
-import { flatMap, isEmpty, isNil, map, merge } from "lodash";
+import { flatMap, groupBy, isEmpty, isNil, map, merge } from "lodash";
```
andreogle, line 8: Would prefer these to import separately as `import flatMap from 'lodash/flatMap';` since this is importing the entire lodash library currently (which is big). I guess it's probably not a major issue for server side applications though

andreogle (commented):
I think this function really needs to be broken up into smaller modules & functions, it's quite difficult to understand what's going on. I'd like to review this in more depth when I get some time

src/start.ts
acenolaza: https://api3dao.atlassian.net/browse/BEC-84

src/start.ts
acenolaza: Right, I'm not sure which option is better at the moment and I'll have to profile it at some point but I'll probably come back to this once Airkeeper is ready and it's time to make improvements.

acenolaza:
> As per comments (which have URLs and detail), you may wish to:
> 
> * look into using Airnode's `go` error and retry library
> * convert the for loop at line 93 to a `map(async ({...}, index) => {...})`
> * make the transaction count `const`
> * make the nonce `const` and `= transactionCount + perSponsorIndex`
> * move the gas retrieval to line ~92 such that it is cached per chain-provider
> 
> Just generally this is very readable and I enjoy your work.

Thanks Aaron for giving such good insights and helpful comments 🥇

src/start.ts
acenolaza: oh nice! Anything that makes the final zip file smaller helps when deploying to aws lambda.

src/start.ts
```diff
@@ -54,265 +54,331 @@ export const handler = async (_event: any = {}): Promise<any> => {
           provider
         );
 
+        const airnodeHDNode = ethers.utils.HDNode.fromMnemonic(
+          nodeSettings.airnodeWalletMnemonic
+        );
+
+        const currentBlock = await provider.getBlockNumber();
```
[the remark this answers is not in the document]
acenolaza: I guess I didn't put it in try/catch block just because I probably copied over from [initialize-provider](https://github.com/api3dao/airnode/blob/6d902da259ec3084c8f4764cadc74e270e5c7162/packages/airnode-node/src/evm/handlers/initialize-provider.ts#L56) in node package but you bring up a good point and since this is the first interaction with the provider if it is down or not reachable then the whole process fails which makes me think that I should put it inside a try/catch.

acenolaza:
> I think this function really needs to be broken up into smaller modules & functions, it's quite difficult to understand what's going on. I'd like to review this in more depth when I get some time

I must say I was expecting this comment from you 😄 but I the main idea was to have something simple that works before the hackathon and then iterate over it until is production ready (or maybe just implement it in Airnode). I would definitely appreciate you looking at Airkeeper in detail at some point.

src/start.ts
acenolaza: Another thing to consider is that we expect to have a single `keeperSponsorWallet` per trigger job meaning that it's a loop because we'd like Airkeeper not to fail but it is expected that this loop runs only once if properly configured.

PR #6, opened 2021-12-23
BEC-84:Switch to Airnode promise-utils for long running async tasks i…

acenolaza:
…n Airkeeper

Async function calls now use the go() function from promise-utils.ts. The operation that go() receives is retryable in all cases with a 5 seconds timeout.

### DIFF
--- a/src/start.ts
+++ b/src/start.ts
@@ -16,16 +16,16 @@ import {
   loadAirkeeperConfig,
   deriveKeeperSponsorWallet,
   printError,
+  retryGo,
 } from "./utils";
 // TODO: use node.evm.getGasPrice() once @api3/airnode-node is updated to v0.4.x
 import { getGasPrice } from "./gas-prices";
 
-export const DEFAULT_RETRY_TIMEOUT_MS = 5_000;
 export const GAS_LIMIT = 500_000;
 
 export const handler = async (_event: any = {}): Promise<any> => {
   const startedAt = new Date();
-  console.log("[DEBUG]\tstarting beaconUpdate...");
+  console.log("[DEBUG]\tstarting Airkeeper...");
   // **************************************************************************
   // 1. Load config (this file must be the same as the one used by the node)
   // **************************************************************************
@@ -69,12 +69,8 @@ export const handler = async (_event: any = {}): Promise<any> => {
         );
 
         // Fetch current block number from chain via provider
-        const retryableGetBlockNumberCall = node.utils.retryOnTimeout(
-          DEFAULT_RETRY_TIMEOUT_MS,
-          () => provider.getBlockNumber()
-        );
-        const [err, currentBlock] = await node.utils.go(
-          () => retryableGetBlockNumberCall
+        const [err, currentBlock] = await retryGo(() =>
+          provider.getBlockNumber()
         );
         if (err || isNil(currentBlock)) {
           printError(err);
@@ -103,16 +99,13 @@ export const handler = async (_event: any = {}): Promise<any> => {
             );
 
             // Fetch keeperSponsorWallet transaction count
-            const retryableGetTransactionCountCall = node.utils.retryOnTimeout(
-              DEFAULT_RETRY_TIMEOUT_MS,
+            const [err, keeperSponsorWalletTransactionCount] = await retryGo(
               () =>
                 provider.getTransactionCount(
                   keeperSponsorWallet.address,
                   currentBlock
                 )
             );
-            const [err, keeperSponsorWalletTransactionCount] =
-              await node.utils.go(() => retryableGetTransactionCountCall);
             if (err || isNil(keeperSponsorWalletTransactionCount)) {
               printError(err);
               console.log(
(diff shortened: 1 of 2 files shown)


src/start.ts
```diff
@@ -67,11 +68,16 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
 
-        let currentBlock: number;
-        try {
-          currentBlock = await provider.getBlockNumber();
-        } catch (error) {
-          printError(error);
+        // Fetch current block number from chain via provider
+        const retryableGetBlockNumberCall = node.utils.retryOnTimeout(
+          DEFAULT_RETRY_TIMEOUT_MS,
+          () => provider.getBlockNumber()
```
aquarat, line 74: Echoing comments I received on my own PR using this library: you can inline this function

src/start.ts
```diff
@@ -67,11 +68,16 @@ export const handler = async (_event: any = {}): Promise<any> => {
           nodeSettings.airnodeWalletMnemonic
         );
 
-        let currentBlock: number;
-        try {
-          currentBlock = await provider.getBlockNumber();
-        } catch (error) {
-          printError(error);
+        // Fetch current block number from chain via provider
+        const retryableGetBlockNumberCall = node.utils.retryOnTimeout(
```
aquarat, line 72: You can use `go()` directly as it uses `retryOnTimeout` underneath.
[`go()`](https://github.com/api3dao/airnode/blob/e0784eb7854b07ca1a159bd44e99bb5d3f8a89fc/packages/airnode-node/src/utils/promise-utils.ts#L30) by default doesn't use a timeout, but you could wrap it to include a timeout in the options object or re-implement it:
Straight use:
```
const [err, currentBlock] = go(provider.getBlockNumber, {timeoutMs: 5000});
```
Wrapped use:
```
[Root scope]
const go = <T>(fn: () => Promise<T>, options?: PromiseOptions) => node.utils.go(fn, {...options, timeoutMs: 5000});

[In the handler function]
// inlined
const [err, currentBlock] = go(provider.getBlockNumber);

// non-inlined due to args
const [err, keeperSponsorWalletTransactionCount] = await go(() =>
                provider.getTransactionCount(
                  keeperSponsorWallet.address,
                  currentBlock
                ));
```
This gives me joy for some reason.

src/start.ts
```diff
@@ -97,11 +103,24 @@ export const handler = async (_event: any = {}): Promise<any> => {
             );
 
             // Fetch keeperSponsorWallet transaction count
-            let keeperSponsorWalletTransactionCount =
-              await provider.getTransactionCount(
-                keeperSponsorWallet.address,
-                currentBlock
+            const retryableGetTransactionCountCall = node.utils.retryOnTimeout(
+              DEFAULT_RETRY_TIMEOUT_MS,
```
aquarat, line 107: You use this so often (10x in this file) that it may be worth wrapping `retryOnTimeout`?
Something like:
```
const retryOnTimeout = <T>(operation: () => Promise<T>, options?: ContinuousRetryOptions) =>
  node.utils.retryOnTimeout(DEFAULT_RETRY_TIMEOUT_MS, operation, options);
```
It's strange that the timeout isn't an extra optional argument.

acenolaza: Good idea. I'll try using the 2nd option

src/start.ts
```diff
@@ -69,12 +69,8 @@ export const handler = async (_event: any = {}): Promise<any> => {
         );
 
         // Fetch current block number from chain via provider
-        const retryableGetBlockNumberCall = node.utils.retryOnTimeout(
-          DEFAULT_RETRY_TIMEOUT_MS,
-          () => provider.getBlockNumber()
-        );
-        const [err, currentBlock] = await node.utils.go(
-          () => retryableGetBlockNumberCall
+        const [err, currentBlock] = await retryGo(() =>
```
aquarat, line 72: `const [err, currentBlock] = await retryGo(provider.getBlockNumber);`
I've had this exact case suggested to me in a PR^
It's looking good! I'm excited :)

aquarat (changes requested):
One small nit, generally it looks great :)

src/start.ts
acenolaza: Yeah. I tried doing that by reading your [first comment](https://github.com/api3dao/airkeeper/pull/6#discussion_r774900426) but unfortunately didn't work as expected.

If I remove the wrapping arrow function then when `getBlockNumber` uses the keyword `this` internally I got a `not defined` error message. I was able to fix it by doing:
```const [err, currentBlock] = await retryGo(provider.getBlockNumber.bind(provider));```
but I thought it was not an improvement over just having a wrapper arrow function.

BTW your suggestion about using the retryGo() function made the code look much nicer 🔥

aquarat: Wow, yeah, interesting. I had a similar thing in my gas pricing PR which was picked up on.

> using the retryGo() function made the code look much nicer

You've done a great job :)

aquarat (approved):
LGTM :fire:

PR #7, opened 2021-12-27
BEC-86: Remove eventLogMaxBlocks from airkeeper.json and use blockHis…

acenolaza:
…toryLimit from config.json instead

The idea is that Airkeeper use the same number of blocks that Airnode uses when fetching history logs from chain.


aquarat (approved):
LGTM :)

PR #8, opened 2022-01-03
BEC-3: Run serverless framework commands from within a docker contain…

acenolaza:
…er for ease of use

With this new Dockerfile, Airkeeper will support deploying and removing the serverless lambda functions without the need for the user to clone, build and install dependencies locally.


README.md
```diff
@@ -8,11 +8,82 @@ Airkeeper will fetch the value from the API (similarly to Airnode) and will also
...
+
+The `deploy` command will create a new AWS lambda function and a new AWS cloud scheduler.
+
+```sh
+docker run -it --rm \
+--env-file config/aws.env \
+--env COMMAND=deploy \
+-v "$(pwd)/config/airkeeper.json:/airkeeper/config/airkeeper.json" \
+-v "$(pwd)/config/config.json:/airkeeper/config/config.json" \
+-v "$(pwd)/config/secrets.env:/airkeeper/config/secrets.env" \
```
amarthadan, line 44: Since all the files are in the same directory that is mapped to the same directory in the container the only volume definition you need is mapping that directory: `"$(pwd)/config:/airkeeper/config"`.
But if you want to stress that these three files are needed it's fine.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    && rm -rf /var/cache/apk/*
+
+RUN git clone --single-branch --branch main https://github.com/api3dao/airkeeper.git
+
+WORKDIR /airkeeper
+
+RUN npm install
+
+CMD npm run sls:config \
+    && npm run sls:$COMMAND
```
amarthadan, line 16: I see multiple issues:
* You're running the command as root. That's usually a bad practice for Docker containers. You should [create a new user and use it to run the commands](https://github.com/api3dao/airnode/blob/07d540f61d3fe29a0213d8cf80046856727b4874/packages/airnode-admin/docker/Dockerfile#L19).
* By running `npm install` you're installing all the development dependencies and the resulting image is huge, more than 700MB
* Using an env variable to provide a command is quite an unusual approach. Usually with the CLI containers (which this essentially is) you use it like any other CLI tool, in this case, it would be for example `docker run ... api3/airkeeper deploy`.

I would suggest installing `serverless` as a global package in the container so that you don't need to install dev dependencies and writing a small `entrypoint.sh` that will pass the correct command from the command line instead of the env variable.

README.md
acenolaza: I initially started running the commands like that since I saw it [here](https://github.com/api3dao/api3-docs/blob/main/docs/airnode/v0.4/grp-providers/docker/deployer-image.md#L80) but then I realized that all the *.example files were also in that directory and I wasn't sure if I should continue to map the whole directory. To be honest I like your suggestion anyways since it's easier for the user to run the commands

Dockerfile
acenolaza: All valid points and I'm going to do some research to start applying those suggestions and I guess I used the env variable for the command because the idea was to come up with something similar to pre-alpha deployer [Dockerfile](https://github.com/api3dao/airnode/blob/pre-alpha/Dockerfile).

acenolaza: These will be resolved by https://api3dao.atlassian.net/browse/BEC-95

PR #9, opened 2022-01-08
Bug fix for local parameters not overriding on-chain template parameters

aquarat:
I may have found a bug, and if it is a bug then this is a fix:

I found that local parameters in the `airkeeper.json` config file weren't overriding on-chain template parameters. This PR fixes the problem - assuming I've understood everything correctly.

### DIFF
--- a/src/start.ts
+++ b/src/start.ts
@@ -155,8 +155,8 @@ export const handler = async (_event: any = {}): Promise<any> => {
                 {}
               );
               const apiCallParameters = {
-                ...configParameters,
                 ...node.evm.encoding.safeDecode(template.parameters),
+                ...configParameters
               };
               const reservedParameters =
                 node.adapters.http.parameters.getReservedParameters(


aquarat:
:)
I discovered this _bug_ through a need/want that arose while I was deploying additional data feeds. I'm currently running four data feeds on 2 chains (Ropsten and Goerli). You can check out the behaviour on the graphs system (it's pretty).

Most of the feeds are price tickers with only the text "eth_usd" as a parameter changing between them. Instead of creating one template per feed I created one template and then overrode a parameter.

On reflection this is probably not good practice because you'd probably want multiple templates (one per pair parameter) to save on gas costs, but I think it would be more useful generally to be able to override parameters in the airkeeper config for templates than compliment existing parameters. This also wouldn't have been a problem if the first template I deployed lacked that parameter.

There's no right or wrong, I just thought I would bring it up for discussion.

bbenligiray:
@aquarat is right, Airkeeper should imitate the node here
https://github.com/api3dao/airnode/blob/e6480fdaa8dc8b86af35d43c61171b02439245e2/packages/airnode-node/src/evm/templates/template-application.ts#L32
The parameters defined in `airkeeper.json` are practically request parameters, so they override template parameters

acenolaza:
I think you might be right tho and whatever params are in the config should have precedence over the params in the template. @bbenligiray any thoughts?

bbenligiray:
@acenolaza See my comment above

PR #11, opened 2022-01-20
BEC-85: Improve log entries in Airkeeper to better differentiate entr…

acenolaza:
…ies that are created asynchronously

Ended up using the logger.ts file from @api3/airnode-node package to have the logs look like the ones seen in Airnode.

image

### DIFF
--- a/src/start.ts
+++ b/src/start.ts
@@ -15,7 +15,6 @@ import { ChainConfig } from "./types";
 import {
   loadAirkeeperConfig,
   deriveKeeperSponsorWallet,
-  printError,
   retryGo,
 } from "./utils";
 // TODO: use node.evm.getGasPrice() once @api3/airnode-node is updated to v0.4.x
@@ -26,7 +25,7 @@ export const BLOCK_COUNT_HISTORY_LIMIT = 300;
 
 export const handler = async (_event: any = {}): Promise<any> => {
   const startedAt = new Date();
-  console.log("[DEBUG]\tstarting Airkeeper...");
+
   // **************************************************************************
   // 1. Load config (this file must be the same as the one used by the node)
   // **************************************************************************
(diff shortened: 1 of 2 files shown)


src/start.ts
```diff
@@ -35,6 +34,14 @@ export const handler = async (_event: any = {}): Promise<any> => {
   const keeperConfig = loadAirkeeperConfig();
   const config = merge(nodeConfig, keeperConfig);
 
+  const baseLogOptions = node.logger.buildBaseOptions(config, {
+    coordinatorId: node.utils.randomString(8),
+  });
+  node.logger.info(
```
aquarat, line 40: This is a minor thing:
You may want to import `logger` itself using the destructured import approach, ala `import {logger} from "@api3/airnode-node";`
This would save a lot of repetition of "node". You can do this with a lot of stuff, like `config`, although that would involve renaming your existing uses of config.

src/start.ts
```diff
@@ -47,11 +54,20 @@ export const handler = async (_event: any = {}): Promise<any> => {
 
   const providerPromises = flatMap(
     evmChains.map((chain: node.ChainConfig & ChainConfig) => {
-      return map(chain.providers, async (chainProvider, _) => {
+      return map(chain.providers, async (chainProvider, providerName) => {
```
aquarat, line 57: This is also a minor issue:
You can actually inline this, ie. remove the 
```
=> {
return 
```
So it becomes 
```
  const providerPromises = flatMap(
    evmChains.map((chain: ChainConfig & ChainConfig) =>
       map(chain.providers, async (chainProvider, providerName) => {
        const providerLogOptions = {
```

src/start.ts
```diff
@@ -289,12 +348,13 @@ export const handler = async (_event: any = {}): Promise<any> => {
               const percentageThreshold =
                 ethers.BigNumber.from(deviationPercentage).mul(basisPoints);
```
aquarat, line 349: There's been a lot of discussion about running deviation thresholds that are less than 1%, I'm not sure if you're familiar with that requirement? You may want to prioritise implementing that next as I they are operating under the assumption that this functionality exists.

aquarat (approved):
In general it looks great, neat as always :)
I've commented on some minor things.

src/start.ts
aquarat: I told Ugur about this and they've created a task to fix it.

src/start.ts
acenolaza: Yes, I had started doing that same thing but Andre pointed out that by doing that we would be importing the whole library. So for consistency I just import everything the same way.

https://github.com/api3dao/airkeeper/pull/5#discussion_r773746721

src/start.ts
acenolaza: I guess here I just wanted to have it look like in [Airnode](https://github.com/api3dao/airnode/blob/869fc03048e344a2efdc2fe5303f31c63db478e1/packages/airnode-node/src/providers/actions.ts#L54)

src/start.ts
aquarat: ah right, sorry about that, thanks :)

PR #12, opened 2022-01-21
BEC-117: Allow for deviation thresholds below 1

acenolaza:
Allowed deviationPercentage lower than 1 but only up to 2 decimal places

### DIFF
--- a/src/start.ts
+++ b/src/start.ts
@@ -160,36 +160,6 @@ export const handler = async (_event: any = {}): Promise<any> => {
               deviationPercentage,
               requestSponsor,
             } of rrpBeaconServerKeeperJobs) {
-              // **************************************************************************
-              // 4. Fetch template by ID
-              // **************************************************************************
-              node.logger.debug(
-                "fetching template...",
-                keeperSponsorWalletLogOptions
-              );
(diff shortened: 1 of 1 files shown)


src/start.ts
```diff
@@ -160,19 +160,43 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+              };
+
+              if (
+                isNaN(Number(deviationPercentage)) ||
+                Number(deviationPercentage) <= 0 ||
+                Number(deviationPercentage) > 100 ||
+                !Number.isInteger(Number(deviationPercentage) * 100)
+              ) {
+                node.logger.error(
+                  `deviationPercentage '${deviationPercentage}' must be a number larger than 0 and less or equal than 100 with no more than 2 decimal places`,
```
andreogle, line 184:
```suggestion
                  `deviationPercentage '${deviationPercentage}' must be a number larger than 0 and less then or equal to 100 with no more than 2 decimal places`,
```

andreogle (approved):

bbenligiray:
For future reference, protocol-v1 PSP conditions for the DapiServer contract assume 10^8 is 100%

PR #13, opened 2022-01-24
BEC-118: Template information in airkeeper.json

acenolaza:
The main idea here is to avoid making calls to AirnodeRrp contract in order to save on blockchain provider calls. The tradeoff is that now the user of Airkeeper must also set each template parameter list in airkeeper.json config.

### DIFF
--- a/src/start.ts
+++ b/src/start.ts
@@ -154,14 +154,14 @@ export const handler = async (_event: any = {}): Promise<any> => {
 
             for (const {
               templateId,
-              parameters,
+              overrideParameters,
               templateParameters,
               oisTitle,
               endpointName,
               deviationPercentage,
               requestSponsor,
             } of rrpBeaconServerKeeperJobs) {
-              const encodedParameters = abi.encode(parameters);
+              const encodedParameters = abi.encode(overrideParameters);
               const beaconId = ethers.utils.solidityKeccak256(
                 ["bytes32", "bytes"],
                 [templateId, encodedParameters]
@@ -229,8 +229,8 @@ export const handler = async (_event: any = {}): Promise<any> => {
                 (e) => e.name === endpointName
               )!;
               const apiCallParameters = [
-                ...parameters,
                 ...templateParameters,
+                ...overrideParameters,
               ].reduce((acc, p) => ({ ...acc, [p.name]: p.value }), {});
               const reservedParameters =
                 node.adapters.http.parameters.getReservedParameters(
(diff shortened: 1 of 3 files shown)


src/start.ts
```diff
@@ -235,21 +221,17 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
-              const configParameters = parameters.reduce(
-                (acc, p) => ({ ...acc, [p.name]: p.value }),
-                {}
-              );
-              const apiCallParameters = {
-                ...node.evm.encoding.safeDecode(template.parameters),
-                ...configParameters,
-              };
+              const apiCallParameters = [
+                ...parameters,
```
aquarat, line 232: This may be the wrong way around: you're overriding override parameters with template parameters?
We generally want configured parameters to override on-chain parameters.

acenolaza: Good catch! I almost introduced the same bug again 😄

andreogle (approved):
Looks good to me, but are we sure this is a tradeoff we want to make?

aquarat:
> Looks good to me, but are we sure this is a tradeoff we want to make?

The plan is to run at least 30 feeds for the hackathon across 4 chains, that's 120 EVM RPC calls per minute just for template retrieval or 5.18 million EVM calls per month (51.8x the Infura free tier limit). 

The worst case scenario related to storing the template parameters locally is a case where they don't match on-chain parameters. @acenolaza has added a check for that by comparing the `templateId` hashes, so in this scenario Airkeeper won't do an update request and this would only happen directly after a faulty deployment.

I think it's a reasonable trade-off (EVM costs vs initial deployment failure potential).

src/start.ts
```diff
@@ -154,13 +154,14 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
-              parameters,
+              overrideParameters,
+              templateParameters,
               oisTitle,
               endpointName,
               deviationPercentage,
               requestSponsor,
             } of rrpBeaconServerKeeperJobs) {
-              const encodedParameters = abi.encode(parameters);
+              const encodedParameters = abi.encode(overrideParameters);
```
aquarat, line 164: This may be incorrect: this output is used in the generation of the `beaconId`, so it should be the sum of all parameters (template and on-chain)?
In most scenarios `overrideParameters` will be an empty array.

aquarat (changes requested):
Looks good, except for one thing :)

src/start.ts
acenolaza: You are correct Aaron. Previously I was getting the whole list of parameters from a single object called parameters and when I refactored into 2 separate arrays is when I made this mistake. Thanks!

aquarat (approved):
LGTM :100:

PR #14, opened 2022-01-26
BEC-128: single api call per rrpBeaconServerKeeperJob

acenolaza:
Requirement for this task is to lower the number of API requests Airkeeper makes by making a single api request per `rrpBeaconServerTriggerJob` without taking into consideration the chain/provider.

I've implemented a similar solution to what @aquarat suggested [here](https://github.com/api3dao/airkeeper/tree/example-cached-api-values). The main difference is that I iterate over the rrpBeaconServerTriggerJobs array outside the chain/provider promises since those run in parallel. The API values are then cached and used when processing each individual provider/keeperSponsorWallet/rrpBeaconServerTriggerJob.

### DIFF
--- a/src/call-api.ts
+++ b/src/call-api.ts
@@ -1,0 +1,153 @@
+import * as abi from "@api3/airnode-abi";
+import * as adapter from "@api3/airnode-adapter";
+import * as node from "@api3/airnode-node";
+import * as ois from "@api3/airnode-ois";
+import * as ethers from "ethers";
+import isNil from "lodash/isNil";
+import { ApiValuesByBeaconId, RrpBeaconServerKeeperTrigger } from "./types";
+import { retryGo } from "./utils";
+
+export const readApiValue = async (
+  airnodeHDNode: ethers.utils.HDNode,
+  oises: ois.OIS[],
+  apiCredentials: node.ApiCredentials[],
+  {
+    oisTitle,
+    endpointName,
+    templateId,
+    templateParameters,
+    overrideParameters,
+  }: RrpBeaconServerKeeperTrigger
+): Promise<node.LogsData<ApiValuesByBeaconId>> => {
+  let apiValue: ethers.BigNumber | null = null;
+  const configParameters = [...templateParameters, ...overrideParameters];
+
+  // Derive beaconId
+  const encodedParameters = abi.encode(configParameters);
+  const beaconId = ethers.utils.solidityKeccak256(
+    ["bytes32", "bytes"],
+    [templateId, encodedParameters]
+  );
+
+  // Verify templateId matches data in rrpBeaconServerKeeperJob
+  const airnodeAddress = airnodeHDNode.derivePath(
+    ethers.utils.defaultPath
+  ).address;
+  const endpointId = ethers.utils.keccak256(
+    ethers.utils.defaultAbiCoder.encode(
+      ["string", "string"],
+      [oisTitle, endpointName]
+    )
+  );
+  const encodedTemplateParameters = abi.encode(templateParameters);
+  const expectedTemplateId = node.evm.templates.getExpectedTemplateId({
+    airnodeAddress,
+    endpointId,
+    encodedParameters: encodedTemplateParameters,
+    id: templateId,
+  });
+  if (expectedTemplateId !== templateId) {
+    return [
+      [
+        node.logger.pend(
+          "ERROR",
+          `templateId '${templateId}' does not match expected templateId '${expectedTemplateId}'`
+        ),
+      ],
+      { [beaconId]: apiValue },
+    ];
+  }
+
+  const configOis = oises.find((o) => o.title === oisTitle)!;
+  const configEndpoint = configOis.endpoints.find(
+    (e) => e.name === endpointName
+  )!;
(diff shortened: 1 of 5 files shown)


src/call-api.ts
```diff
@@ -0,0 +1,153 @@
...
+  apiCredentials: node.ApiCredentials[],
+  {
+    oisTitle,
+    endpointName,
+    templateId,
+    templateParameters,
+    overrideParameters,
+  }: RrpBeaconServerKeeperTrigger
+): Promise<node.LogsData<ApiValuesByBeaconId>> => {
+  let apiValue: ethers.BigNumber | null = null;
```
aquarat, line 22: You probably saw this coming, but... now that you've moved all this into a function you can remove the let declaration. It's not a biggy of course.

The value of apiValue is always null, until after a successful call to the API in your try below, so all the other instances can be replaced with null and the successful response can be immediately returned.

```
  const encodedTemplateParameters = abi.encode(templateParameters);
  const expectedTemplateId = node.evm.templates.getExpectedTemplateId({
    airnodeAddress,
    endpointId,
    encodedParameters: encodedTemplateParameters,
    id: templateId,
  });
  if (expectedTemplateId !== templateId) {
    return [
      [
        node.logger.pend(
          "ERROR",
          `templateId '${templateId}' does not match expected templateId '${expectedTemplateId}'`
        ),
      ],
      { [beaconId]: null },
    ];
  }
```

```
  // Extract API value
  try {
    const response = adapter.extractAndEncodeResponse(
      apiResponse!.data,
      reservedParameters as adapter.ReservedParameters
    );
    const apiValue = ethers.BigNumber.from(response.values[0].toString());

    const logApiValue = node.logger.pend(
      "INFO",
      `API value: ${apiValue.toString()}`
    );

    return [[logApiResponse, logApiValue], { [beaconId]: apiValue }]
  } catch (error) {
    return [
      [
        node.logger.pend(
          "ERROR",
          `failed to extract or encode value from API response: ${JSON.stringify(
            apiResponse.data
          )}`,
          error as any
        ),
      ],
      { [beaconId]: null },
    ];
  }
```

aquarat (approved):
It looks great to me, except for one thing :stuck_out_tongue: 
Good catch on the concurrency.

src/call-api.ts
```diff
@@ -0,0 +1,152 @@
+import * as abi from "@api3/airnode-abi";
+import * as adapter from "@api3/airnode-adapter";
+import * as node from "@api3/airnode-node";
+import * as ois from "@api3/airnode-ois";
+import * as ethers from "ethers";
```
andreogle, line 5: Not sure if this makes a difference though but it seems a bit more consistent?
```suggestion
import { ethers } from "ethers";
```

src/call-api.ts
```diff
@@ -0,0 +1,152 @@
...
+    node.adapters.http.parameters.getReservedParameters(
+      configEndpoint,
+      apiCallParameters || {}
+    );
+  if (!reservedParameters._type) {
+    return [
+      [
+        node.logger.pend(
+          "ERROR",
+          `reserved parameter 'type' is missing for endpoint: ${endpointName}`
```
andreogle, line 78:
```suggestion
          `reserved parameter '_type' is missing for endpoint: ${endpointName}`
```

andreogle: Nit: maybe we could save quite a bit of space in a few places with something like:

```ts
const message = `reserved parameter '_type' is missing for endpoint: ${endpointName}`;
return [[log], { [beaconId]: null }];
```

3 lines vs 8 lines

src/call-api.ts
```diff
@@ -0,0 +1,152 @@
...
+          "ERROR",
+          `failed to fetch data from API for endpoint: ${endpointName}`,
+          errBuildAndExecuteRequest
+        ),
+      ],
+      { [beaconId]: null },
+    ];
+  }
+  const logApiResponse = node.logger.pend(
+    "INFO",
```
andreogle, line 121: Should this maybe be DEBUG instead? These can typically get quite large

src/start.ts
```diff
@@ -43,6 +42,49 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+  );
+
+  const apiValuePromises = triggers.rrpBeaconServerKeeperJobs.map((job) =>
+    retryGo(() => readApiValue(airnodeHDNode, oises, apiCredentials, job))
+  );
+  const responses = await Promise.all(apiValuePromises);
+
+  // Group logs and API values by beaconId
+  const logsAndApiValuesByBeaconId: {
+    [beaconId: string]: {
```
andreogle, line 62: Maybe this type should be defined elsewhere?

andreogle (approved):

src/call-api.ts
acenolaza: Wouldn't that look a bit weird in files like this one: https://github.com/api3dao/airkeeper/blob/bec128-api-calls/src/types.ts#L0-L1

I'm 💯 in favor of consistency but I'm not so sure how to achieve it with the imports tbh 😄

acenolaza: Maybe I should just do:
```suggestion
import ethers from "ethers";
```

PR #15, opened 2022-01-27
BEC-129: Add eslint/prettier/husky to Airkeeper

acenolaza:
Jira Issue: https://api3dao.atlassian.net/browse/BEC-129


aquarat (approved):
This is great, makes me happier than it should :)

PR #17, opened 2022-01-29
Revert "Merge pull request #15 from api3dao/bec129-eslint-prettier-hu…

acenolaza:
…sky"

This reverts commit 4f10eb627f0bd0b4f6c2a2d8fec6e3d6ad862c65, reversing
changes made to ad2568d4fed72a68950003a493092f16b33b392c.

Unfortunately I need to revert this because something got broken when trying to deploy to aws the lambda function. Locally it works but when executing `npm run sls:deploy` the deployed lambda function fails to run with a "Cannot find module" error message. I believe it is because I added `noEmit=true` to tsconfig.json that causes .build directory to not have the proper js files built and copied in there. In the meantime I'd like to revert the merge commit and properly test the change prior to committing it again.


aquarat (approved):
LGTM

PR #18, opened 2022-01-31
Fixes context for function being passed to go() when trying to get th…

acenolaza:
…e gas price for legacy chains

Same fix applied in airnode-node package  https://github.com/api3dao/airnode/pull/799


aquarat (approved):
LGTM :)

PR #22, opened 2022-02-07
BEC-95: Docker build improvements

amarthadan:
Changes made:
* Removed `serverless-plugin-typescript` plugin as its usage in the container made the container slower and bigger. I'm instead building the sources manually.
* There is no need for `sls config credentials` in the container if you make the credentials available in standard places (either config file or env variables)


acenolaza (approved):
Works like a charm 🏆

PR #24, opened 2022-02-07
Adds unit tests for call-api

acenolaza:
Adds jest and tests around call-api.ts


src/call-api.test.ts
```diff
@@ -0,0 +1,394 @@
...
+            {
+              name: "amount",
+              default: "1",
+              operationParameter: {
+                name: "amount",
+                in: "query",
+              },
+            },
+          ],
+          testable: true,
```
amarthadan, line 105: Not sure what version is this for but this no longer belongs to the OSI section in `v0.4.x`.

src/call-api.test.ts
```diff
@@ -0,0 +1,394 @@
...
+          level: "ERROR",
+          message: expect.stringMatching(
+            "reserved parameter '_type' is missing for endpoint: convertToUSD"
+          ),
+        },
+      ])
+    );
+    expect(Object.values(res)).toEqual([null]);
+    expect(spy).not.toHaveBeenCalled();
+  });
```
amarthadan, line 286: What's the difference between this test and the previous one?

amarthadan (changes requested):
Since you added tests it would be nice to set up CI so they run for every PR.

src/call-api.test.ts
acenolaza: Airkeeper currently only works with v0.3.1 and we'll probably update it's dependencies after the ETH Denver hackathon.

src/call-api.test.ts
acenolaza: The call-api.ts in Airkeeper is similar but not a copy paste from call-api.ts in Airnode and its only function receives different arguments. I know it might look redundant at first but since this is a different app I think it needs to have basic testing. This function might even change in the near future and I would be safer to refactor with tests in place.

amarthadan: No, you misunderstood, I'm totally fine with having tests, it just seems that you have copied the same test twice, the `returns null if reserved parameter '_type' is missing` one.

acenolaza: Right, sorry about that. It was not easy to see it in the diff. It seems I just duplicated that code accidentally. I'll remove it

.github/workflows/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone airkeeper
+        uses: actions/checkout@master
```
amarthadan, line 16: I'm not that familiar with GitHub actions but why do you use `actions/checkout@master` here and `actions/checkout@v2` for the `compile` job?

amarthadan (approved):
I'm not sure but I'd say that the current config will install the dependencies twice, once for `compile` and once for `test`. But I might be wrong. Either way, it's probably not an issue right now.

.github/workflows/main.yml
acenolaza: good catch! It was a bad copy/paste I guess and it seems that the right value is `v2` https://github.com/actions/checkout

acenolaza:
> I'm not sure but I'd say that the current config will install the dependencies twice, once for `compile` and once for `test`. But I might be wrong. Either way, it's probably not an issue right now.

I'll need to do some research on github actions before I can answer this.

acenolaza:
> > I'm not sure but I'd say that the current config will install the dependencies twice, once for `compile` and once for `test`. But I might be wrong. Either way, it's probably not an issue right now.
> 
> I'll need to do some research on github actions before I can answer this.

It seems that each job runs on a fresh environment but It also seems that there's a way to cache these dependencies to speed up the workflow https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#caching-dependencies

amarthadan:
> > > I'm not sure but I'd say that the current config will install the dependencies twice, once for `compile` and once for `test`. But I might be wrong. Either way, it's probably not an issue right now.
> > 
> > 
> > I'll need to do some research on github actions before I can answer this.
> 
> It seems that each job runs on a fresh environment but It also seems that there's a way to cache these dependencies to speed up the workflow https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#caching-dependencies

I think you can use `actions/setup-node@v2`'s [in-built caching](https://github.com/actions/setup-node#caching-packages-dependencies) by just adding `cache: 'yarn'`. Not really sure if it's that simple, I know that ChainAPI guys were doing something like that.

acenolaza:
> > > > I'm not sure but I'd say that the current config will install the dependencies twice, once for `compile` and once for `test`. But I might be wrong. Either way, it's probably not an issue right now.
> > > 
> > > 
> > > I'll need to do some research on github actions before I can answer this.
> > 
> > 
> > It seems that each job runs on a fresh environment but It also seems that there's a way to cache these dependencies to speed up the workflow https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#caching-dependencies
> 
> I think you can use `actions/setup-node@v2`'s [in-built caching](https://github.com/actions/setup-node#caching-packages-dependencies) by just adding `cache: 'yarn'`. Not really sure if it's that simple, I know that ChainAPI guys were doing something like that.

using `cache: "yarn'"` reduces from 22/23 secs to 8/9 secs the `Install dependencies` step 🥳

Siegrift:
Can we also do this for Airnode and other project such as dao dashboard? Or maybe create a task? @dcroote did you try this when you looked for AIrnode CI optimizations?

dcroote:
> I think you can use actions/setup-node@v2's [in-built caching](https://github.com/actions/setup-node#caching-packages-dependencies) by just adding cache: 'yarn'. Not really sure if it's that simple, I know that ChainAPI guys were doing something like that.

In this case it is actually is that simple 👍 

@Siegrift  - because Airnode is a monorepo it's a bit more complicated, but I did implement caching a few months ago here: https://github.com/api3dao/airnode/pull/663.

PR #25, opened 2022-02-08
Bec129 eslint prettier husky

acenolaza:


.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn lint && yarn tsc
```
aquarat, line 4: Perhaps this should be in `pre-commit`? or maybe pre-push isn't necessary?
The reason I ask is because you'd ideally want it to lint/warn about issues prior to committing.

.prettierignore
```diff
@@ -0,0 +1,4 @@
+# Ignore artifacts:
+build
```
aquarat, line 2: SLS creates a folder called `.build` which isn't included, you may want to include it (as prettier does try and frisk it) :smile:

aquarat (approved):
LGTM :smile: but @Siegrift should also check as he's better with this.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
amarthadan, line 4: Why use `npx` instead of using `lint-staged` directly from development dependencies?

.prettierignore
amarthadan: Same goes for all the `config/*.json` files

amarthadan (approved):

.husky/pre-push
acenolaza: The idea was to run prettier --write on pre-commit and prettier --check on pre-push but I see your point 🤔

.husky/pre-commit
acenolaza: Good question. I just followed this guide from the prettier docs https://prettier.io/docs/en/install.html#git-hooks

.husky/pre-push
acenolaza: I basically copied this pattern from [api3-dao-dashboard](https://github.com/api3dao/api3-dao-dashboard/blob/main/.husky/pre-push) so maybe @Siegrift can help out here.

amarthadan: Please, do not run `prettier --write` on pre-commit or anywhere else for that matter automatically. The developers should be always aware of the changes they're committing.

acenolaza: I'm confused 🤔 Isn't Airnode and all other projects doing exactly that already? https://github.com/api3dao/airnode/blob/master/.husky/pre-commit

amarthadan: That's only for the validator autogenerated docs. For the rest it's calling `yarn lint` and that's calling `yarn prettier:check` which is just a check showing the errors, not fixing them automatically.

acenolaza: Right, but there are still multiple other repos where we are doing `--write` like in [operations](https://github.com/api3dao/operations/blob/main/.husky/pre-commit#L3), [services](https://github.com/api3dao/services/blob/main/.husky/pre-commit#L0-L1), etc. I wonder if we should make this consistent at across repos.

amarthadan: I think we should and IMO it's not a good idea to have an automatic write. But I don't want to force it unless others agree.

acenolaza: Removed --write and pre-push hock

amarthadan (approved):

PR #26, opened 2022-02-10
Removed mnemonic dependency

Ashar2shahid:
I've added `airnode` and `airnodeXpub` to the `airkeeper.json`. I've placed them on the root of the json because airkeeper is designed to be used with only one airnode. Also adjusted the types but not entirely sure if it was the correct way of doing it.

### DIFF
--- a/src/start.ts
+++ b/src/start.ts
@@ -35,7 +35,7 @@ export const handler = async (_event: any = {}): Promise<any> => {
   });
   node.logger.info(`Airkeeper started at ${node.utils.formatDateTime(startedAt)}`, baseLogOptions);
 
-  const { chains: keeperChains, triggers: keeperTriggers } = keeperConfig;
+  const {  airnode: airnodeAddress, airnodeXpub: airnodeXpub ,chains: keeperChains, triggers: keeperTriggers } = keeperConfig;
   const config = {
     ...nodeConfig,
     chains: keeperChains.map((chain) => {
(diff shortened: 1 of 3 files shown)


src/start.ts
```diff
@@ -50,18 +50,20 @@ export const handler = async (_event: any = {}): Promise<any> => {
     }),
     triggers: { ...nodeConfig.triggers, ...keeperTriggers },
   };
-  const { chains, nodeSettings, triggers, ois: oises, apiCredentials } = config;
+  const { chains, triggers, ois: oises, apiCredentials } = config;
+
+  const airnodeHDNode = ethers.utils.HDNode.fromExtendedKey(airnodeXpub);
+  if (airnodeAddress !== airnodeHDNode.derivePath('0/0').address) {
+    throw new Error(`xpub does not belong to Airnode: ${airnodeAddress}`);
+  }
```
acenolaza, line 58: Nice! I see this is the same code form airnode-admin-cli [verifyAirnodeAddress()](https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-admin/src/implementation.ts#L41) function 👌🏻

src/start.ts
```diff
@@ -35,7 +35,12 @@ export const handler = async (_event: any = {}): Promise<any> => {
   });
   node.logger.info(`Airkeeper started at ${node.utils.formatDateTime(startedAt)}`, baseLogOptions);
 
-  const {  airnode: airnodeAddress, airnodeXpub: airnodeXpub ,chains: keeperChains, triggers: keeperTriggers } = keeperConfig;
+  const {
+    airnode: airnodeAddress,
+    airnodeXpub: airnodeXpub,
```
aquarat, line 40: Hey, well done on doing this. I think you can remove the `: airnodeXpub`, so it's just 
```
const {  airnode: airnodeAddress, airnodeXpub,chains: keeperChains, triggers: keeperTriggers } = keeperConfig;
```
It's generally better to not map variable names if you don't need to, so if there's a way of keeping airnode = airnode, chains = chain, etc. then you can just do 
```
const {  airnode, airnodeXpub, chains, triggers } = keeperConfig;
```
which is easier to read/less confusing :)

aquarat (changes requested):
Very cool that you've done this Ashar :)
I suggest you remove the variable mappings as per my only comment to make it easier to read/less complex.

src/start.ts
Ashar2shahid: thanks for the suggestion, The xpub is needed to derive the sponsorWallets so I think I have to keep it there.

> It's generally better to not map variable names if you don't need to, so if there's a way of keeping airnode = airnode, chains = chain, etc. then you can just do

ah okay, well I didn't want to disturb santiago's code as much as possible but i'll do the other suggestion.

aquarat: If you're busy with the conference I can make the changes? This would be useful for the backup Amberdata Airkeeper.

Let me know. I'll probably still need assistance with the sponsor wallets.

Ashar2shahid: Yeah go ahead, im travelling today so won't be able to get to it until in the evening

acenolaza (approved):
🏆

PR #28, opened 2022-02-17
Review of PSP WIP

bbenligiray:


psp_notes.md
```diff
@@ -0,0 +1,86 @@
...
+6. Using the response from the API, the Airnode checks if the condition specified in the `conditions` field of the
+   subscription is met (in most cases, by making a static call to a function)
+
+I'm confused about
+[this](https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol#L20-L23)
+step because this comment says that it is not needed for beacon updates.
+
+If I still need to check the condition in Airkeeper then I guess I need to get the percentage value used when calling
+DapiServer.registerBeaconUpdateSubscription() and lastly call DapiServer.conditionPspBeaconUpdate(). The percentage
+value should be in the config but also in the subscription details from step 2.
```
bbenligiray, line 70: The percentage value should only be in the subscription details, it shouldn't be separately specified in the config file

Here's an example of how you would do that
https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/test/dapis/DapiServer.sol.js#L136
but I think we should do
```js
beaconUpdateSubscriptionConditions = hre.ethers.utils.defaultAbiCoder.encode(
    ['bytes32', 'bytes32', 'bytes32', 'bytes32', 'bytes'],
    [
      hre.ethers.utils.formatBytes32String('1bB'),
      hre.ethers.utils.formatBytes32String('_conditionFunctionId'),
      hre.ethers.utils.defaultAbiCoder.encode(
        ['bytes4'],
        [dapiServer.interface.getSighash('conditionPspBeaconUpdate')]
      ),
      hre.ethers.utils.formatBytes32String('_conditionParameters'),
      beaconUpdateSubscriptionConditionParameters,
    ]
  );
```
and the lack of `_conditionChainId` means you should use chain the requester contract is on and the lack of `_conditionAddress` means you should use the requester contract address (so that `conditions` isn't huge)

psp_notes.md
```diff
@@ -0,0 +1,86 @@
...
+
+7. If the condition is met, the Airnode fulfills the subscription by calling the specified function using the sponsor
+   wallet
+
+   This step means that Airkeeper will need to derive the sponsor wallet, use the airnode wallet to sign the message (or
+   should we use relayer here?) and call DapiServer.fulfillPspBeaconUpdate().
```
bbenligiray, line 76: Airnode and the relayer are the same so it doesn't matter. Note that the signature scheme changed so you should be signing the sponsor wallet address and not the data.

psp_notes.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Other questions:
+
+1. Should Airkeeper run the new coordinator lambda function every minute like current lamdba function?
```
bbenligiray, line 80: Yes

psp_notes.md
```diff
@@ -0,0 +1,86 @@
...
+1. What was the requirement for having the need of more than one lambda function and a coordinator function? Answer in
+   51:35 https://drive.google.com/file/d/14BaWqZojFCy7ZQrnp8V2sxDP-mGyLdGV/view
```
bbenligiray, line 82: It's mainly to be able to scale up the computational power so that you can derive a bunch of sponsor wallets and sign a bunch of transactions. If you had a single coordinator it wouldn't be able to maintain 500 beacons within 60 seconds (?)

psp_notes.md
```diff
@@ -0,0 +1,86 @@
...
+1. How to test this??? Should I try adding these new contracts to airnode-operations package and set everything there so
+   I can run my tests locally?
```
bbenligiray, line 84: I would just implement a hardhat script in the airkeeper repo that sets up the contracts you need at the local node and not involve the monorepo to this at all

psp_notes.md
```diff
@@ -0,0 +1,86 @@
...
+1. Operations repo will probably generate all the data for the template and subscriptions but does it have to actually
+   store it onchain?
```
bbenligiray, line 86: No, as mentioned before

psp_notes.md
bbenligiray: The initial task was to do everything in the coordinator anyway so we can test this out

config/psp.example.json
```diff
@@ -0,0 +1,69 @@
+{
```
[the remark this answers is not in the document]
bbenligiray: I think the format should be
```json
{
  "chains": [...],
  "triggers": {
    "proto-psp": [
      "SUBSCRIPTION ID #1 GOES HERE",
      "SUBSCRIPTION ID #2 GOES HERE",
      ...
    ]
  },
  "subscriptions": {
    "SUBSCRIPTION ID #1 GOES HERE": {
      "SUBSCRIPTION ID #1 FIELDS GO HERE"
    },
    "SUBSCRIPTION ID #2 GOES HERE": {
      "SUBSCRIPTION ID #2 FIELDS GO HERE"
    },
    ...
  },
  "templates": {
    "TEMPLATE ID #1 GOES HERE": {
      "TEMPLATE ID #1 FIELDS GO HERE"
    },
    "TEMPLATE ID #2 GOES HERE": {
      "TEMPLATE ID #2 FIELDS GO HERE"
    },
    ...
  }
}
```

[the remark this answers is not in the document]
bbenligiray: Here, proto-psp means the PSP implementation that doesn't care about allocators, authorizers, sponsorships. See the README comments for more info.

psp_notes.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. Deploy AirnodeProtocol contract
+1. Deploy AccessControlRegistry or AccessControlRegistryAdminned contracts
+1. Deploy DapiServer contract
+1. Deploy AllocatorWithAirnode or AllocatorWithManager contracts
+1. **Do we need any whitelisting on DapiServer contract for readers or others?** It seems that reader == address(0) was
+   kept so Airkeeper should still be able to read without the need to whitelist. But now that I think about it the
+   condition function is the one reading the value therefore Airkeeper might not even need to read the current data
+   point value in order to determine if update is needed.
+1. Derive sponsor wallet, fund it and call AirnodeProtocol.setPspSponsorshipStatus()
```
bbenligiray, line 11: Skip setting the sponsorship status

config/psp.example.json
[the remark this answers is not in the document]
bbenligiray: Probably a better option to the above is to specify which subscription ID is supposed to be served by which slot. So it would go like
```json
{
  ...,
  "triggers": {
    "proto-psp": {
      "1": "SUBSCRIPTION ID #1 GOES HERE",
      "2": "SUBSCRIPTION ID #2 GOES HERE",
      ...
    }
  },
  ...
}
```
where `"1"` and `"2"` represent the slot indices. This would allow us to use allocators and hardcoded subscriptions simultaneously.
Btw I'm not sure if we should skip `0` as a slot index. We probably should.

config/psp.example.json
```diff
@@ -0,0 +1,69 @@
...
+        "DapiServer": "0xFF298a3169117606BDA78eD6C04917224968f9b5"
+      },
+      "allocators": [
+        {
+          "address": "0xFF298a3169117606BDA78eD6C04917224968f9b5",
+          "startIndex": "0",
+          "endIndex": "5"
+        }
+      ]
+    },
```
bbenligiray, line 16: Looks good to me (for when we'll use allocators). Alternatively we may want to have a list of indices for a bit more flexibility.

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+
+        // **************************************************************************
+        // Initialize provider specific data
+        // **************************************************************************
+        node.logger.debug('Initializing provider...', providerLogOptions);
+
+        const chainProviderUrl = chainProvider.url || '';
+        const provider = node.evm.buildEVMProvider(chainProviderUrl, chain.id);
+
+        //TODO: where to get abi from?
```
bbenligiray, line 100: From the respective `deployments/` directory when we'll have that

bbenligiray: I will be figuring this out

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+        const subscriptionIds: string[] = [];
+        for (const { address, startIndex, endIndex } of chain.allocators) {
+          const allocator = new ethers.Contract(address, allocatorAbi, provider);
+          const staticCalldatas = [];
+          for (let i = startIndex; i <= endIndex; i++) {
+            staticCalldatas.push(
+              allocator.interface.encodeFunctionData('airnodeToSlotIndexToSlot', [airnodeAddress, i])
+            );
+          }
+          // TODO: retryGo?
```
bbenligiray, line 277: It feels like retrying is overrated

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+          resultDatas.forEach((resultData: ethers.utils.BytesLike) => {
+            const result = allocator.interface.decodeFunctionResult('airnodeToSlotIndexToSlot', resultData);
+            if (result.subscriptionId && result.expirationTimestamp.gt(nowInSeconds())) {
+              subscriptionIds.push(result.subscriptionId);
+            }
+          });
+        }
+
+        // **************************************************************************
+        // Process each psp subscription in serial to keep nonces in order
```
bbenligiray, line 288: You should only process subscriptions serially (sorted by ID) if they use the same sponsor

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+          // Fetch subscription details
+          // **************************************************************************
+          node.logger.debug('Fetching subscription details...', subscriptionIdLogOptions);
+
+          // On-chain details: AirnodeProtocol.subscriptions(id)
+          // Off-chain details: triggers.psp
+          const subscription = triggers.psp.find((s) => {
+            // TODO: Should the config contain subscriptionId field
+            // or should we derive it using the other fields to verify
+            // that values in config are correct?
```
bbenligiray, line 308: The config file will include the subscription/template ID, the node use these IDs to get the respective details, then recalculate the ID with the details to verify it

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+            continue;
+          }
+
+          // **************************************************************************
+          // Check authorization status
+          // **************************************************************************
+          node.logger.debug('Checking authorization status...', subscriptionIdLogOptions);
+
+          // TODO: This needs a little bit of thinking 🤔
+          // By default subscription.requester is set to the dapiServer address when calling dapiServer.registerBeaconUpdateSubscription()
```
bbenligiray, line 332: That's why checking the authorization doesn't make much sense, the Airnode operator probably wants to authorize the DapiServer contract

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+            node.logger.warn('API value is missing. Skipping update...', subscriptionIdLogOptions);
+            continue;
+          }
+
+          // **************************************************************************
+          // Check conditions
+          // **************************************************************************
+          node.logger.debug('Checking conditions...', subscriptionIdLogOptions);
+
+          // Is this step really needed?
```
bbenligiray, line 414: Yes

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+          // https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol#L20-L23
+
+          const encodedFulfillmentData = ethers.utils.defaultAbiCoder.encode(['int256'], [apiValue]);
+
+          // TODO: Should "subscription.conditions" be already encoded in config?
+          // const encodedConditionParameters = ethers.utils.defaultAbiCoder.encode(
+          //   ['uint256'],
+          //   [subscription.conditions] // How is this supposed to be set in config? i.e. (10**18 / 10) for 10%?
+          // );
```
bbenligiray, line 423: Yeah, I think you should decode `conditions` using airnode-abi each time you need to do the condition check to get `_conditionParameters`

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+            continue;
+          }
+          node.logger.info('Conditions met. Updating beacon...', subscriptionIdLogOptions);
+
+          // **************************************************************************
+          // Compute signature
+          // **************************************************************************
+          node.logger.debug('Signing fulfill message...', subscriptionIdLogOptions);
+
+          const airnodeWallet = ethers.Wallet.fromMnemonic(nodeSettings.airnodeWalletMnemonic).connect(provider);
```
bbenligiray, line 442: Do this once globally because this is probably computationally intensive. You shouldn't need to connect the provider to sign a message so this is chain/provider agnostic.

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+          // TODO: protocol parameter is missing in v0.3.1
+          // const sponsorWallet = node.evm.deriveSponsorWallet(airnodeHDNode, subscription.sponsor).connect(provider);
+          const sponsorWallet = deriveSponsorWallet(
+            nodeSettings.airnodeWalletMnemonic,
+            subscription.sponsor,
+            '2'
```
bbenligiray, line 448: Use 3 here, 2 is relayed RRP (?) I don't know which makes more sense though, this is debatable

src/psp.ts
```diff
@@ -0,0 +1,502 @@
...
+              )
+            )
+          );
+
+          // **************************************************************************
+          // Update beacon
+          // **************************************************************************
+          node.logger.debug('Fulfilling subscription...', subscriptionIdLogOptions);
+
+          // TODO: what if there is a previous pending fulfillPspBeaconUpdate (or any pending?) tx for this sponsor wallet?
```
bbenligiray, line 469: You should derive sponsor wallet addresses at the beginning and get the transaction counts immediately. Then you make the fulfillment transactions at the end by incrementing that (like Airnode). This means repeated fulfillments will have the same nonce and cancel each other out.

src/psp.ts
acenolaza: I still need to add the check for validating the subscriptionId matches the subscription details

src/psp.ts
acenolaza: To keep it simple at first I just register the subscription with `conditions` being the threshold value but I guess this can become more complicated and `airnode-abi` will make more sense to be used.

src/psp.ts
acenolaza: Oh right, I always forget about setting the nonce. I'm already doing it for current Airkeeper implementation: https://github.com/api3dao/airkeeper/blob/psp-beacon-update/src/start.ts#L161-L176

acenolaza (commented):
Thanks for the feedback @bbenligiray!

src/psp.ts
acenolaza: Is there really a value to sort them by subscriptionId?

src/psp.ts
bbenligiray: That would definitely work in this specific case but proper airnode-abi encoding will allow us to use proto-psp with contracts other than DapiServer

src/psp.ts
acenolaza: Sure thing 👍🏻

acenolaza:
> @acenolaza let me know when you need a second pass

Yesterday I've pushed a commit with most (if not all) suggestions you've made but there are still some TODOs in the psp.ts file that need to be resolved.

acenolaza:
@bbenligiray please review these changes again if you can. Thanks

PR #29, opened 2022-02-24
 Review of PSP WIP v2

bbenligiray:


psp_notes.md
```diff
@@ -0,0 +1,86 @@
+## Steps required befor PSP flow can be started:
```
bbenligiray, line 1: Can you update this file and ask for a review again please?

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+        if (errorGetBlockNumber || isNil(currentBlock)) {
+          node.logger.error('Failed to fetch the blockNumber', {
+            ...providerLogOptions,
+            error: errorGetBlockNumber,
+          });
+          return;
+        }
+
+        /*
+        //TODO: where to get abi from?
```
bbenligiray, line 92: Remove this

bbenligiray, line 92: Remove this

bbenligiray, line 92: Remove this

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+        /* 
+        // **************************************************************************
+        // Fetch subscriptions from allocators
+        // DISABLED: https://github.com/api3dao/airkeeper/pull/28#discussion_r808586658
+        // **************************************************************************
+        node.logger.debug('Fetching subcriptions...', providerLogOptions);
+
+        
+        const subscriptionIds: string[] = [];
+        //TODO: where to get abi from?
```
bbenligiray, line 210: Remove this

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+          // **************************************************************************
+          // Derive sponsorWallet address
+          // **************************************************************************
+          node.logger.debug('Deriving sponsorWallet...', providerLogOptions);
+
+          // TODO: switch to node.evm.deriveSponsorWallet when @api3/airnode-node allows setting the `protocolId`
+          const sponsorWallet = deriveSponsorWallet(
+            nodeSettings.airnodeWalletMnemonic,
+            sponsor,
+            '3' // TODO: should this be in a centralized enum somewhere (api3/airnode-protocol maybe)?
```
bbenligiray, line 356: People prefer 2 here

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+            node.logger.error('Failed to fetch the sponsor wallet transaction count', {
+              ...sponsorWalletLogOptions,
+              error: errorGetTransactionCount,
+            });
+            return;
+          }
+          let nonce = sponsorWalletTransactionCount;
+
+          // **************************************************************************
+          // Process each psp subscription in serial to keep nonces in order
```
bbenligiray, line 384: I'm not sure if this has to be serial. We can talk about that at the call.

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+          // **************************************************************************
+          node.logger.debug('Processing rrpBeaconServerKeeperJobs...', sponsorWalletLogOptions);
```
bbenligiray, line 386: rrpBeaconServer -> dapiServer

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+          } of sponsorPspTriggersWithSubscriptions || []) {
+            const subscriptionIdLogOptions = {
+              ...sponsorWalletLogOptions,
+              additional: {
+                ...sponsorWalletLogOptions.additional,
+                subscriptionId,
+              },
+            };
+
+            /*
```
bbenligiray, line 407: You can remove these imo. They will not come back in production.

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+
+            const encodedFulfillmentData = ethers.utils.defaultAbiCoder.encode(['int256'], [apiValue]);
+            let conditionFunction: ethers.utils.FunctionFragment;
+            let conditionParameters: string;
+            try {
+              const decodedConditions = abi.decode(conditions);
+              const [decodedConditionFunctionId] = ethers.utils.defaultAbiCoder.decode(
+                ['bytes4'],
+                decodedConditions._conditionFunctionId
+              );
```
bbenligiray, line 535: Note that Airnode ABI only supports `bytes32`. So I'd decode `bytes32` and get rid of the leading zeros manually to get the `bytes4`.

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+              });
+              continue;
+            }
+            if (!conditionsMet) {
+              node.logger.warn('Conditions not met. Skipping update...', subscriptionIdLogOptions);
+              continue;
+            }
+
+            // **************************************************************************
+            // Fetch current gas fee data
```
bbenligiray, line 563: We probably can get away with doing this only once while getting the block number

src/psp.ts
```diff
@@ -0,0 +1,664 @@
...
+                  {
+                    gasLimit: GAS_LIMIT,
+                    ...gasTarget,
+                    nonce: nonce++,
+                  }
+                )
+            );
+            if (errfulfillFunction) {
+              node.logger.error(
+                `failed to submit transaction using wallet ${sponsorWallet.address} with nonce ${currentNonce}`,
```
bbenligiray, line 633: Uses the incremented nonce

config/airkeeper.example.json
```diff
@@ -5,30 +5,56 @@
     {
       "id": "31337",
       "contracts": {
-        "RrpBeaconServer": "0x610178dA211FEF7D417bC0e6FeD39F05609AD788"
+        "RrpBeaconServer": "0x610178dA211FEF7D417bC0e6FeD39F05609AD788",
+        "AirnodeProtocol": "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512",
```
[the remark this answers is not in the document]
acenolaza: Right, that's a left over from when I was checking the sponsorship status

config/airkeeper.example.json
```diff
@@ -5,30 +5,56 @@
...
         "endpointName": "convertToUSD",
         "deviationPercentage": "5",
         "keeperSponsor": "0x2479808b1216E998309A727df8A0A98A1130A162",
         "requestSponsor": "0x61648B2Ec3e6b3492E90184Ef281C2ba28a675ec"
       }
+    ],
+    "proto-psp": [
+      {
+        "subscriptionId": "0xdf6d2785e1fc71ba3aa08fccb77d74ac4ba678208bc33c445152ea999a0cc8af",
+        "overrideParameters": [{ "type": "string32", "name": "from", "value": "ETH" }],
```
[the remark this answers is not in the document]
acenolaza: This is a mistake. I tried to apply the same logic used for previous rrp beacon update where I could reuse a template and still derive a different `beaconId` but now that we are no longer using the parameters to derive the beaconId I guess we must create a template per beacon/subscription.

config/airkeeper.example.json
```diff
@@ -5,30 +5,56 @@
...
+        "oisTitle": "Currency Converter API",
+        "endpointName": "convertToUSD"
```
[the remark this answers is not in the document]
acenolaza: Right I was struggling with placing these 2 values and ended up replicating oisTitle and endpointName just to be consistent with airnode config.json rrp triggers.

I'll add the endpoints object in the root of the file but since this is a breaking change for old rrp beacon update function, should we consider renaming `rrpBeaconServerKeeperJobs` or some other property in the airkeeper.json config?

config/airkeeper.example.json
```diff
@@ -5,30 +5,56 @@
...
+  },
+  "templates": {
+    "0x6f737bbf31dfed584a16f53b7d725ff64bee67e79a468259456fb40aa19c60c4": {
+      "endpointId": "0x13dea3311fe0d6b84f4daeab831befbc49e19e6494c41e9e065a09c3c68f43b6",
+      "templateParameters": [
+        { "type": "string32", "name": "to", "value": "USD" },
+        { "type": "string32", "name": "_type", "value": "int256" },
+        { "type": "string32", "name": "_path", "value": "result" },
+        { "type": "string32", "name": "_times", "value": "1000000" }
+      ]
```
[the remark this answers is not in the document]
acenolaza: I can do that but the thing is that I tried to put the exact same fields that would be saved on-chain and in the case we remove them from here switching to those values instead of reading them from config would be easier. Are you thinking that we could use the encoded values directly to speed things up? is this field in subscription be ever used?

psp_notes.md
acenolaza: I was planning to just delete it after properly updating the README.

src/psp.ts
acenolaza: Is this how I'm supposed to be encoding the `conditionFunctionId`?:

```
  const beaconUpdateSubscriptionConditions = [
    {
      type: 'bytes32',
      name: '_conditionFunctionId',
      value: hre.ethers.utils.defaultAbiCoder.encode(
        ['bytes4'],
        dapiServer.interface.getSighash('conditionPspBeaconUpdate')
      ),
    },
    { type: 'bytes', name: '_conditionParameters', value: beaconUpdateSubscriptionConditionParameters },
  ];
```
If so then I could be sure that the encoded string within the `bytes32` is a `bytes4`, right?

config/airkeeper.example.json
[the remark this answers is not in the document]
bbenligiray: I don't think that we have to worry about breaking Airkeeper

config/airkeeper.example.json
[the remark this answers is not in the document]
bbenligiray: You will have to encode these to verify the template ID anyway so keeping them decoded doesn't actually save time. I'm just saying that either both should be decoded or both should be encoded. I prefer encoded because it's less ambiguous, people may try to change the order of decoded parameters for example and that would change the template/subscription IDs.

src/psp.ts
bbenligiray: Right

src/psp.ts
acenolaza: This might be a bit misleading but in reality the fulfill tx is submitted with the value of `currentNonce` because 'nonce++' is incrementing the value after assignment. Maybe I should separate this into a new line to make it more explicit and easier to read.

bbenligiray: Yeah I misread his. However, it feels like this won't work when you make multiple transactions and increment `nonce` multiple times.

PR #31, opened 2022-03-09
BEC-210: refactor airkeeper

acenolaza:
I took a similar approach to Airnode with it's coordinator managing all the steps. Now it should be easier to setup spawning new lambdas after this separation into new functions.


src/evm/initialize-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ethers } from 'ethers';
+import isNil from 'lodash/isNil';
+import { ChainConfig, EVMProviderState } from '../types';
+import { retryGo } from '../utils';
+
+const rrpBeaconServerAbi = new ethers.utils.Interface(protocol.RrpBeaconServerFactory.abi).format(
+  ethers.utils.FormatTypes.minimal
+);
+
+//TODO: where to get abi from?
```
aquarat, line 13: If the contract always implements these two signatures then you shouldn't need to retrieve the ABI - this fragment will be fine I think?

src/evm/initialize-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+  DapiServer: dapiServerAbi,
+};
+
+export const initializeProvider = async (
+  chain: ChainConfig,
+  providerUrl: string
+): Promise<LogsData<EVMProviderState | null>> => {
+  const provider = node.evm.buildEVMProvider(providerUrl, chain.id);
+
+  const contracts = Object.entries(chain.contracts).reduce((acc, [contractName, contractAddress]) => {
```
aquarat, line 30: Nice :+1:

src/evm/process-transaction.ts
```diff
@@ -0,0 +1,75 @@
...
+  airnodeWallet: ethers.Wallet,
+  contract: ethers.Contract,
+  gasTarget: node.GasTarget,
+  subscriptions: ProcessableSubscription[],
+  sponsorWallet: ethers.Wallet
+): Promise<node.LogsData<ProcessableSubscription>[]> => {
+  const logs: node.LogsData<ProcessableSubscription>[] = [];
+
+  // Process each subscription in serial to keep nonces in order
+  for (const subscription of subscriptions.sort((a, b) => a.nonce - b.nonce) || []) {
```
aquarat, line 17: I feel like I read this exact comment somewhere else:
Are you sure these need to be done serially?
While working on another task recently I routinely did 7 transactions asynchronously with the nonces incremented in a `.map` and I never received any errors. You should be able to convert this for loop to a map and the return from the map via `Promise.all` could be flattened into `logs`, which would remove the type and initialisation. Lmk if you disagree, once again I'm curious.

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+
+const initializeState = (config: Config): State => {
+  const { triggers, subscriptions } = config;
+
+  const baseLogOptions = node.logger.buildBaseOptions(config, {
+    coordinatorId: node.utils.randomHexString(8),
+  });
+
+  const enabledSubscriptions: Id<Subscription>[] = [];
+  triggers['proto-psp'].forEach((subscriptionId) => {
```
aquarat, line 63: This can be turned into a `map`, I kind of did it already:
```
  const enabledSubscriptions = config.triggers['proto-psp']
    .filter((subscriptionId) => config.subscriptions[subscriptionId])
    .map((subscriptionId) => ({
      ...config.subscriptions[subscriptionId],
      id: subscriptionId,
    }))
    .filter((subscription) => {
      const subscriptionId = subscription.id;
      // Get subscriptions details
      if (isNil(subscription)) {
        node.logger.warn(`SubscriptionId ${subscriptionId} not found in subscriptions`, baseLogOptions);
        return;
      }
      // Verify subscriptionId
      const expectedSubscriptionId = ethers.utils.solidityKeccak256(
        ['uint256', 'address', 'bytes32', 'bytes', 'bytes', 'address', 'address', 'address', 'bytes4'],
        [
          subscription.chainId,
          subscription.airnodeAddress,
          subscription.templateId,
          subscription.parameters,
          subscription.conditions,
          subscription.relayer,
          subscription.sponsor,
          subscription.requester,
          subscription.fulfillFunctionId,
        ]
      );
      if (subscriptionId !== expectedSubscriptionId) {
        node.logger.warn(
          `SubscriptionId ${subscriptionId} does not match expected ${expectedSubscriptionId}`,
          baseLogOptions
        );
        return;
      }

      return true;
    });

  if (isEmpty(enabledSubscriptions)) {
    node.logger.info('No proto-psp subscriptions to process', baseLogOptions);

    return {
      config,
      baseLogOptions,
      groupedSubscriptions: [],
      apiValuesBySubscriptionId: {},
      providerStates: [],
    };
  }
```

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+    groupedSubscriptions,
+    apiValuesBySubscriptionId: {},
+    providerStates: [],
+  };
+};
+
+const executeApiCalls = async (state: State): Promise<State> => {
+  const { config, baseLogOptions, groupedSubscriptions } = state;
+
+  // TODO: promise.all? 🤔
```
aquarat, line 159: ;) It'll work well here to do this as `Promise.all` as it'll make the apiCalls async, which are I/O heavy (would greatly benefit from this behaviour).

I have a suspicion that it is unnecessary to split out callApi into a separate Lambda. Only crypto operations are compute-heavy enough to warrant being split out.

```
  const promisedApiValuesBySubscriptionId = groupedSubscriptions
    .map(async ({ subscriptions, template, endpoint }) => {
    const templateLogOptions: node.LogOptions = {
      ...baseLogOptions,
      additional: {
        templateId: template.id,
      },
    };
```

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+
+  return { ...state, apiValuesBySubscriptionId };
+};
+
+const initializeProviders = async (state: State): Promise<State> => {
+  const { config, baseLogOptions } = state;
+
+  const airnodeWallet = ethers.Wallet.fromMnemonic(config.nodeSettings.airnodeWalletMnemonic);
+
+  const evmChains = config.chains.filter((chain: ChainConfig) => chain.type === 'evm');
```
aquarat, line 204: Just fyi you don't need the type:
```
const evmChains = config.chains.filter(chain => chain.type === 'evm');
```

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+  if (isEmpty(evmChains)) {
+    throw new Error('One or more evm compatible chain(s) must be defined in the provided config');
```
aquarat, line 206: You can remove the (): `One or more evm compatible chains must be defined in the provided config`

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+  }
+  const providerPromises = flatMap(
```
aquarat, line 208: There is a `flatMap` array function that could replace the lodash `flatMap`. `flatMap` is native so may be more performant. You can also remove the types and replace lodash `map` with the native type: (I think `Object.entries` is a new thing)
```
  const providerPromises =
    evmChains.flatMap(chain =>
      Object.entries(chain.providers).map(async ([providerName, chainProvider]) => {
        const providerLogOptions = {
          ...baseLogOptions,
          meta: {
            ...baseLogOptions.meta,
            chainId: chain.id,
            providerName,
          },
        };
```

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+          node.logger.warn('Failed to initialize provider', providerLogOptions);
+          return null;
+        }
+
+        return {
+          airnodeWallet,
+          chainId: chain.id,
+          providerName,
+          ...evmProviderState,
+        } as ProviderState<EVMProviderState>;
```
aquarat, line 233: It looks like this type can also be removed.

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+      })
+    )
+  );
+
+  const providerStates = await Promise.all(providerPromises);
```
aquarat, line 238: :+1: :muscle: :rocket:

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+    const subscriptionLogOptions: node.LogOptions = {
+      ...logOptions,
+      additional: {
+        ...logOptions.additional,
+        subscriptionId: data.subscription.id,
+      },
+    };
+    node.logger.logPending(log, subscriptionLogOptions);
+    if (data.isValid) {
+      validSubscriptions.push({
```
aquarat, line 269: You can avoid mutation by using `.filter`, eg. :
```
  const conditionPromises = subscriptions.map(
    (subscription) =>
      checkSubscriptionCondition(subscription, apiValuesBySubscriptionId[subscription.id], contract, voidSigner).then(
        ([logs, isValid]) => [logs, {subscription, isValid}]
      ) as Promise<node.LogsData<{ subscription: Id<Subscription>; isValid: boolean }>>
  );
  const result = await Promise.all(conditionPromises);

  return result.filter(([_log, data]) => data.isValid).map(([log, data]) => {
    const subscriptionLogOptions: node.LogOptions = {
      ...logOptions,
      additional: {
        ...logOptions.additional,
        subscriptionId: data.subscription.id,
      },
    };
    node.logger.logPending(log, subscriptionLogOptions);
    return {
      ...data.subscription,
      apiValue: apiValuesBySubscriptionId[data.subscription.id],
    };
  });
```

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+};
+
+const groupSubscriptionsBySponsorWallet = async (
+  subscriptionsBySponsor: Dictionary<CheckedSubscription[]>,
+  config: Config,
+  provider: ethers.providers.Provider,
+  currentBlock: number,
+  providerLogOptions: node.LogOptions
+): Promise<SponsorWalletWithSubscriptions[]> => {
+  const sponsorWalletsWithSubscriptions: SponsorWalletWithSubscriptions[] = [];
```
aquarat, line 286: I can't help myself:
```
  const sponsorAddresses = Object.keys(subscriptionsBySponsor);
  const sponsorWalletAndTransactionCountPromises = sponsorAddresses.map(
    (sponsor) =>
      getSponsorWalletAndTransactionCount(config, provider, currentBlock, sponsor).then(([logs, data]) => [
        logs,
        {...data, sponsor},
      ]) as Promise<node.LogsData<(SponsorWalletTransactionCount | null) & { sponsor: string }>>
  );
  const sponsorWalletsAndTransactionCounts = await Promise.all(sponsorWalletAndTransactionCountPromises);

  const makeSponsorLogOptions = (extra: {}) => ({
    ...providerLogOptions,
    additional: {
      ...providerLogOptions.additional,
      ...extra
    }
  });

  return sponsorWalletsAndTransactionCounts.filter(([_logs, data]) => {
    if (isNil(data.sponsorWallet) || isNil(data.transactionCount)) {
      node.logger.warn('Failed to fetch sponsor wallet or transaction count', makeSponsorLogOptions({sponsor: data.sponsor}));
      return;
    }

    return true;
  })
    .map(([logs, data]) => {
      node.logger.logPending(logs, makeSponsorLogOptions({sponsor: data.sponsor}));

      return {
        subscriptions: subscriptionsBySponsor[data.sponsor].map((subscription) => ({
          ...subscription,
          nonce: data.transactionCount,
        })),
        sponsorWallet: data.sponsorWallet,
      };
    });
};
```

aquarat: Those log options are horrible... perhaps you could have a set of factory(?) functions that allow you to compose them more easily with less duplication (like above).

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+    });
+  });
+
+  return sponsorWalletsWithSubscriptions;
+};
+
+const submitTransactions = async (state: State) => {
+  const { config, baseLogOptions, groupedSubscriptions, apiValuesBySubscriptionId, providerStates } = state;
+
+  const providerPromises = providerStates.map(async (providerState) => {
```
aquarat, line 327: Delicious :smile:

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+      ...baseLogOptions,
+      meta: {
+        ...baseLogOptions.meta,
+        chainId,
+        providerName,
+      },
+    };
+
+    // Get subscriptions from template/endnpoint groups
+    const subscriptions = flatMap(groupedSubscriptions.map((s) => s.subscriptions));
```
aquarat, line 341:
```
const subscriptions = groupedSubscriptions.flatMap(s => s.subscriptions);
```

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+  });
+
+  await Promise.all(providerPromises);
+};
+
+const updateBeacon = async (config: Config) => {
+  // =================================================================
+  // STEP 1: Initialize state
+  // =================================================================
+  let state: State = initializeState(config);
```
aquarat, line 414: ;)

aquarat (commented):
Some constructive comments. As always your work is excellent :)

serverless.yml
```diff
@@ -18,12 +18,12 @@ package:
...
 functions:
   rrp-beacon-update:
-    handler: dist/src/rrp.beaconUpdate
+    handler: dist/src/handlers/rrp.handler
     timeout: 60
     events:
       - schedule: rate(1 minute)
   psp-beacon-update:
-    handler: dist/src/psp.beaconUpdate
+    handler: dist/src/handlers/psp.handler
```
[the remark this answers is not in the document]
amarthadan: We were using that but got rid of it because it was causing the Docker image to be unnecessarily big since it needed TypeScript as a run dependency.

src/evm/check-conditions.ts
```diff
@@ -0,0 +1,56 @@
...
+      .connect(voidSigner)
+      .functions[conditionFunction.name](subscription.id, encodedFulfillmentData, conditionParameters)
+  );
+  if (errorConditionFunction || isNil(result)) {
+    const message = 'Failed to check conditions';
+    const log = node.logger.pend('ERROR', message, errorConditionFunction);
+    return [[log], false];
+  }
+  // The result will always be ethers.Result type even if solidity function retuns a single value
+  // See https://docs.ethers.io/v5/api/contract/contract/#Contract-functionsCall
```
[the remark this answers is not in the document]
amarthadan: I think this is the second case a bit lower
`contract.functions.METHOD_NAME( ...args [ , overrides ] ) => Promise< [Result](https://docs.ethers.io/v5/api/utils/abi/interface/#Result) >`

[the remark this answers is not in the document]
acenolaza: Always trust the docs over my comments 😄

src/evm/initialize-provider.ts
acenolaza: Yes, I don't expect this to change but that's just human-readable ABI. Still I get the feeling that this should be coming from the protocol or services package 🤔

src/evm/process-transaction.ts
acenolaza: > I feel like I read this exact comment somewhere else:

probably [here](https://github.com/api3dao/airkeeper/blob/bec210-refactor-airkeeper/src/handlers/rrp.ts#L209) 😁 

> Are you sure these need to be done serially?

Not 💯  sure but it when writing the RRP beacon update my concern was that if a tx gets rejected or something then the nonce order would get messed up.

> While working on another task recently I routinely did 7 transactions asynchronously with the nonces incremented in a .map and I never received any errors. You should be able to convert this for loop to a map and the return from the map via Promise.all could be flattened into logs, which would remove the type and initialisation. Lmk if you disagree, once again I'm curious.

I will definitely test this and if it works then I would prefer to run them all in parallel.

src/handlers/psp.ts
acenolaza: Yes, I was expecting to refactor this into a promise.all(). If you have the fix and have already tested it please feel free to push it. Otherwise I'll do since I know how busy you are these days.

src/handlers/psp.ts
acenolaza: With Michal we were wondering if we could reuse the callApi() function from the airnode-node package.  Maybe that would be enough and we don't even need to start a new function lambda for making api calls if it's not a bottleneck.

src/handlers/psp.ts
acenolaza: I guess I did it because airnode-node is also [checking](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/providers/actions.ts#L47) the chain type but yes. It's probably not necessary.

src/handlers/psp.ts
acenolaza: I was also wondering why we use the lodash functions when there are native functions for this but I didn't want to run into any weird issues so I just did it the same way it's done in [airnode-node](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/providers/actions.ts#L53)

src/handlers/psp.ts
acenolaza: I almost implemented exactly that 🤯 but then I thought that the `forEach` approach was easier to read. I don't feel too strong about this so I might as well use your implementation 💪🏻

src/handlers/psp.ts
acenolaza: Indeed they are 😆  and they add a bunch of lines to the code but the end result is really nice:

image

I'm going to think about how to improve it. Open to suggestions

src/handlers/psp.ts
acenolaza: Should I just rename this function  to `state()` and pass an arg called `state` 🤣

serverless.yml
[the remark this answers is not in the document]
aquarat: All good, I just wanted to understand, thank you :)

src/evm/process-transaction.ts
aquarat: Is there a scenario where a middle transaction would fail but the preceding tx and following txs would/would be able to succeed? 

This is assuming the chain services/permissions have been correctly configured.

src/handlers/psp.ts
aquarat: I'm not so busy now, but I tested most of these in my IDE before commenting them :smile: 
Sometime my brain won't leave things alone :/

src/handlers/psp.ts
aquarat: That sounds pretty cool :)

src/handlers/psp.ts
aquarat: No no, I meant you don't need `(chain: ChainConfig)` only `chain` ^
I can't comment on the actual check, but I'd imagine it is indeed unnecessary.

src/handlers/psp.ts
aquarat: `airnode-node` is far from perfect :stuck_out_tongue: 
I did some more research and found that lodash functions are 2-3x slower than native but only for _reference_ types, like Object. Types like `number` and `boolean` are faster for lodash weirdly.

src/handlers/psp.ts
aquarat: I'm just trying to keep Andre happy ;)

src/handlers/psp.ts
aquarat: Sorry, my comment may have come across severely.
I support you using the existing logging solution, in fact I think it's very cool that you did, but I suggest you use a function to help you compose the logOptions so that there's less duplication. I see there's an open Jira task to make it so that the log subsystem is configured up-front instead of with every call.

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
+import * as abi from '@api3/airnode-abi';
+import * as node from '@api3/airnode-node';
+import { ethers } from 'ethers';
+import { Dictionary } from 'lodash';
+import flatMap from 'lodash/flatMap';
```
aquarat, line 5: btw, you can do this here:
```
import {flatMap, isEmpty, isNil } from 'lodash';
```

src/api/call-api.ts
```diff
@@ -15,21 +15,15 @@ export const callApi = async ({
...
     const log = node.logger.pend('ERROR', message);
     return [[log], null];
   }
 
   // Remove reserved parameters
-  const sanitizedParameters: adapter.Parameters = node.utils.removeKeys(
-    apiCallParameters || {},
-    ois.RESERVED_PARAMETERS
-  );
+  const sanitizedParameters: adapter.Parameters = node.utils.removeKeys(apiCallParameters, ois.RESERVED_PARAMETERS);
```
andreogle, line 26: Is `apiCallParameters` always going to exist? I see the `|| {}` default has been removed

src/config.ts
```diff
@@ -0,0 +1,48 @@
...
+    ...airnodeConfig,
+    chains: airkeeperConfig.chains.map((chain) => {
+      if (isNil(chain.id)) {
+        throw new Error(`Missing 'id' property in chain config: ${JSON.stringify(chain)}`);
+      }
+      const configChain = airnodeConfig.chains.find((c) => c.id === chain.id);
+      if (isNil(configChain)) {
+        throw new Error(`Chain id ${chain.id} not found in node config.json`);
+      }
+      return merge(configChain, chain);
```
andreogle, line 39: Why not just  spread here?

src/constants.ts
```diff
@@ -1,3 +1,14 @@
-export const GAS_LIMIT = 500_000;
+// The number of past blocks to lookup when fetching events.
 export const BLOCK_COUNT_HISTORY_LIMIT = 300;
+
+// The default amount of time before a "retryable" promise is timed out and retried
 export const DEFAULT_RETRY_TIMEOUT_MS = 5_000;
+
+// The Priority Fee in Wei
+export const PRIORITY_FEE = 3120000000;
```
andreogle, line 8: There was a [change recently](https://github.com/api3dao/airnode/pull/933/files#diff-09bfa05deaf24c684349763bbc069a051da261e1a972b10df5d6ca252abcf5e9) to rename this to `PRIORITY_FEE_IN_WEI`. We might want to replicate that here

```suggestion
export const PRIORITY_FEE_IN_WEI = 3_120_000_000;
```

src/evm/initialize-provider.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  // Fetch current block number
+  const [errorGetBlockNumber, currentBlock] = await retryGo(() => provider.getBlockNumber());
+  if (errorGetBlockNumber || isNil(currentBlock)) {
+    const message = 'Failed to fetch the blockNumber';
+    const log = node.logger.pend('ERROR', message, errorGetBlockNumber);
+    return [[log], null];
+  }
+  const currentBlockMessage = `Current block number for chainId ${chain.id}: ${currentBlock}`;
+  const currentBlockLog = node.logger.pend('DEBUG', currentBlockMessage);
```
andreogle, line 45: Maybe this should be `INFO` instead?

src/evm/initialize-provider.ts
```diff
@@ -0,0 +1,70 @@
...
+    provider,
+    chainOptions: chain.options,
+  });
+  if (!gasTarget) {
+    const message = 'Failed to fetch gas price';
+    const log = node.logger.pend('ERROR', message);
+    return [[log], null];
+  }
+  const gasTargetMessage = `Gas target for chainId ${chain.id}: ${JSON.stringify(gasTarget)}`;
+  const gasTargetLog = node.logger.pend('DEBUG', gasTargetMessage);
```
andreogle, line 58: Same with this

src/evm/transaction-count.ts
```diff
@@ -0,0 +1,38 @@
...
+  provider: ethers.providers.Provider,
+  currentBlock: number,
+  sponsor: string
+): Promise<node.LogsData<SponsorWalletTransactionCount | null>> => {
+  // Derive sponsorWallet address
+  // TODO: switch to node.evm.deriveSponsorWallet when @api3/airnode-node allows setting the `protocolId`
+  const sponsorWallet = deriveSponsorWallet(
+    airnodeWallet.mnemonic.phrase,
+    sponsor,
+    '2' // TODO: should this be in a centralized enum somewhere (api3/airnode-protocol maybe)?
```
andreogle, line 19: What is this number even for? It's not obvious at all so a descriptive variable name would help a lot

src/handlers/psp.ts
```diff
@@ -0,0 +1,436 @@
...
+      additional: {
+        templateId: template.id,
+      },
+    };
+    const apiCallParameters = abi.decode(template.templateParameters);
+    const [errorCallApi, logsData] = await retryGo(() =>
+      callApi({
+        oises: config.ois,
+        apiCredentials: config.apiCredentials,
+        apiCallParameters,
```
aquarat, line 173: You do the API Call with `templateParameters`, but you also have parameters in Airkeeper's config (subscription.parameters) and it looks like those parameters are only used to get the `expectedSubscriptionId`? Are they the same parameters? Is this intentional?

src/evm/transaction-count.ts
```diff
@@ -0,0 +1,38 @@
...
+  const [errorGetTransactionCount, transactionCount] = await retryGo(() =>
+    provider.getTransactionCount(sponsorWallet.address, currentBlock)
+  );
+  if (errorGetTransactionCount || isNil(transactionCount)) {
+    const message = 'Failed to fetch the sponsor wallet transaction count';
+    const log = node.logger.pend('ERROR', message, errorGetTransactionCount);
+    return [[log], null];
+  }
+
+  const message = `Sponsor wallet ${sponsorWallet.address.replace(
```
andreogle, line 32: Maybe these interpolated function calls should be done separately? This is kinda difficult to read

src/utils.ts
```diff
@@ -1,48 +1,7 @@
...
-    airnodeMnemonic,
-    `m/44'/60'/0'/${deriveWalletPathFromSponsorAddress(sponsorAddress, protocolId)}`
-  );
-}
-
 const retryGo = <T>(fn: () => Promise<T>, options?: node.utils.PromiseOptions) =>
   node.utils.go(() => node.utils.retryOnTimeout(DEFAULT_RETRY_TIMEOUT_MS, fn), options);
 
-export { loadNodeConfig, parseConfig, deriveSponsorWallet, retryGo };
+export { retryGo };
```
andreogle, line 7: Why not just inline the `export` with the function declaration?

andreogle: Seems like we're mixing export styles in a few places?

src/handlers/psp.ts
andreogle: We've generally tried specifically NOT to do this as lodash doesn't have proper tree shaking (so you get the entire library)

amarthadan:
Even though not all the comments were resolved I would suggest merging this and continuing from there as this is kinda blocking all the other tasks. WDYT?

src/evm/process-transaction.ts
acenolaza: 🤔 not sure to be honest, maybe a timeout from a provider or something totally unexpected like that.

The thing is that we probably don't want to keep submitting txs if for instance sponsor wallet runs out of funds. My thinking was that as soon as a tx reverts or fails then we stop sending txs with this wallet.

src/handlers/psp.ts
acenolaza: One thing to notice tho is that I want to call `node.logger.logPending()` for all elements in `results` and not just for those whose `data.isValid` is equal to `true`

src/handlers/psp.ts
acenolaza: Subscriptions seems to allow parameters to be set [here](https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/protocol/StorageUtils.sol#L29) but they are being hardcoded to empty string [here](https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol#L274). Maybe @bbenligiray can better explain the use-case for `subscription.parameters`.

src/evm/transaction-count.ts
aquarat: FYI It's the protocol version :)

src/config.ts
aquarat: Both objects contain a "contracts" object, you'd want to merge "contracts" rather than overwrite?

aquarat: Still, I guess that could be done with:
```
return {...configChain, ...chain, contracts: {...configChain.contracts, ...chain.contracts}};
```

src/handlers/psp.ts
aquarat: Noted, sorry, will avoid :)

src/api/call-api.ts
acenolaza: good question 🤔 

let's see, `abi.decode` doesn't return null or undefined and worst case is empty object: https://github.com/api3dao/airkeeper/blob/bec210-refactor-airkeeper/src/handlers/psp.ts#L168

`CallApiOptions` interface also doesn't allow null or undefined.

I guess that if endpoint doesn't use any parameter then this argument should be an empty object making the `|| {}` a bit redundant.

Do you still think is worth adding that extra safety measure?

src/config.ts
acenolaza: Becuase I needed deep merge to be able to also merge `chains.contracts` enumerable https://lodash.com/docs/#merge

src/evm/transaction-count.ts
acenolaza: This is the protocol ID part of the derivation path. We use 1 for RRP, 2 for PSP and 12345 for old Airkeeper beacon updater lambda. 

More info on sponsor wallet derivation path scheme: https://github.com/api3dao/airnode/blob/master/packages/airnode-admin/src/implementation.ts#L95-L109

src/utils.ts
acenolaza: I'll update the export style 👍🏻

src/api/call-api.ts
andreogle: It depends on what behaviour we want here. If something has gone wrong (that shouldn't have gone wrong), then we should probably let this throw/crash?

amarthadan (approved):
> Most of the comments were resolved I think now it is in a pretty good state to merge

In that case, let's do it.

PR #33, opened 2022-03-11
airkeeper.json config validator

vponline:
This adds a Zod schema and validator for the `airkeeper.json` configuration file.

The schema is based on `airkeeper.example.json` and assumes that all fields there are required.


src/validator/validator.ts
```diff
@@ -0,0 +1,71 @@
...
+  airnodeXpub: z.string(),
+  chains: z.array(chainConfigSchema),
+  triggers: triggersSchema,
+  subscriptions: subscriptionsSchema,
+  templates: templatesSchema,
+  endpoints: endpointsSchema,
+});
+
+export type SchemaType<Schema extends ZodFirstPartySchemaTypes> = z.infer<Schema>;
+export type Config = SchemaType<typeof configSchema>;
```
vponline, line 69: Wondering if `Config` is too general here and should be something like `AirkeeperConfig` to avoid confusing it with Airnode `Config`?

vponline, line 69: Same here regarding `configSchema`

acenolaza:
I like these changes but I still wonder if we should have implemented them out of [this](https://github.com/api3dao/airkeeper/tree/bec210-refactor-airkeeper) branch since I've done a small refactor around config parsing and loading.

vponline:
> I like these changes but I still wonder if we should have implemented them out of [this](https://github.com/api3dao/airkeeper/tree/bec210-refactor-airkeeper) branch since I've done a small refactor around config parsing and loading.

You're right that makes more sense, I'll update this PR and switch it to that branch.

src/config.ts
```diff
@@ -20,12 +21,24 @@ const loadAirnodeConfig = () => {
...
+  return validationOutput.data;
+};
+
 const parseConfig = <T>(filename: string): T => {
   const configPath = path.resolve(__dirname, '..', '..', 'config', `${filename}.json`);
   return JSON.parse(fs.readFileSync(configPath, 'utf8'));
 };
 
-const mergeConfigs = (airnodeConfig: node.Config, airkeeperConfig: Config): Config => {
+const mergeConfigs = (airnodeConfig: node.Config, airkeeperConfig: AirkeeperConfig): Config => {
```
amarthadan, line 41: Can you change the type of `Config` (the resulting merged one) to reflect the new types? I mean use the ones generated by zod.

amarthadan (changes requested):
Looks good, just some minor tweaks.

src/validator/validator.ts
amarthadan: It is a bit confusing, especially since we have the `Config` type in `types.ts` representing the merged config. I think you can rename it as you're suggesting.

config/airkeeper.example.json
```diff
@@ -28,7 +28,7 @@
         "requestSponsor": "0x61648B2Ec3e6b3492E90184Ef281C2ba28a675ec"
       }
     ],
-    "proto-psp": ["0x168194af62ab1b621eff3be1df9646f198dcef36f9eace0474fd19d47b2e0039"]
+    "protoPsp": ["0x168194af62ab1b621eff3be1df9646f198dcef36f9eace0474fd19d47b2e0039"]
```
aquarat, line 31: I like this.

src/config.ts
```diff
@@ -20,12 +21,19 @@ export const loadAirnodeConfig = () => {
   return config;
 };
 
-export const parseConfig = <T>(filename: string): T => {
-  const configPath = path.resolve(__dirname, '..', '..', 'config', `${filename}.json`);
-  return JSON.parse(fs.readFileSync(configPath, 'utf8'));
+export const loadAirkeeperConfig = (): AirkeeperConfig => {
```
aquarat, line 24: You don't need this type definition: `AirkeeperConfig`

src/config.ts
```diff
@@ -20,12 +21,19 @@ export const loadAirnodeConfig = () => {
...
+  const configPath = path.resolve(__dirname, '..', '..', 'config', `airkeeper.json`);
+  const airkeeperConfig: AirkeeperConfig = JSON.parse(fs.readFileSync(configPath, 'utf8'));
```
aquarat, line 26: You don't need this type definition either: `AirkeeperConfig`

src/config.ts
```diff
@@ -20,12 +21,19 @@ export const loadAirnodeConfig = () => {
...
+  const validationOutput = validateConfig(airkeeperConfig);
+  if (!validationOutput.success) {
+    throw new Error(`Invalid Airkeeper configuration file: ${JSON.stringify(validationOutput.error, null, 2)}`);
+  }
+
+  return validationOutput.data;
 };
 
-export const mergeConfigs = (airnodeConfig: node.Config, airkeeperConfig: Config): Config => {
+export const mergeConfigs = (airnodeConfig: node.Config, airkeeperConfig: AirkeeperConfig): Config => {
```
aquarat, line 36: You don't need this type definition: `Config`

src/handlers/psp.ts
```diff
@@ -96,7 +96,7 @@ const initializeState = (config: Config): State => {
 
   const groupedSubscriptions: GroupedSubscriptions[] = [];
   if (isEmpty(enabledSubscriptions)) {
-    node.logger.info('No proto-psp subscriptions to process', baseLogOptions);
+    node.logger.info('No protoPsp subscriptions to process', baseLogOptions);
```
aquarat, line 99: Minor, dunno if this should perhaps be "proto-PSP".

src/validator/validator.test.ts
```diff
@@ -0,0 +1,91 @@
+import { readFileSync } from 'fs';
```
aquarat, line 1: Nice @ tests

aquarat (approved):
Only minor notes, great work, well done :)

src/validator/validator.ts
```diff
@@ -0,0 +1,83 @@
...
+  subscriptions: subscriptionsSchema,
+  templates: templatesSchema,
+  endpoints: endpointsSchema,
+});
+
+export type SchemaType<Schema extends ZodFirstPartySchemaTypes> = z.infer<Schema>;
+export type AirkeeperConfig = SchemaType<typeof configSchema>;
+export type AirkeeperChainContracts = z.infer<typeof chainContractsSchema>;
+export type AirkeeperChainConfig = z.infer<typeof chainSchema>;
+export type Trigger = z.infer<typeof triggerSchema>;
```
aquarat, line 74: This doesn't appear to be used.

amarthadan: I don't see that as a problem. I'm quite fond of having types for all the parts of the config if I need them.

amarthadan (approved):

src/validator/validator.ts
vponline: I've fixed the other points @aquarat, so I'll merge this and we can remove this at a later time if it is not used by then :smile:

PR #34, opened 2022-03-15
Fix broken CI by fixing a broken URL

aquarat:
This fixes the CI - it was a broken URL.
The document itself might be out of data, but it's probably a valuable reference regardless.

### DIFF
--- a/psp_notes.md
+++ b/psp_notes.md
@@ -55,8 +55,9 @@ https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-node/src/ev
 
 5. The Airnode makes the API call specified by the template and the additional parameters
 
-Probably current [call-api-ts](https://github.com/api3dao/airkeeper/blob/main/src/call-api.ts) in Airkeeper can be
-re-used for this step.
+Probably current
+[call-api-ts](https://github.com/api3dao/airkeeper/blob/2de24dd82d710d1dc383049f2637a57b15d90ae2/src/call-api.ts) in
+Airkeeper can be re-used for this step.
 
 6. Using the response from the API, the Airnode checks if the condition specified in the `conditions` field of the
    subscription is met (in most cases, by making a static call to a function)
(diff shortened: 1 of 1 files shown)


psp_notes.md
```diff
@@ -55,8 +55,9 @@ https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-node/src/ev
 
 5. The Airnode makes the API call specified by the template and the additional parameters
 
-Probably current [call-api-ts](https://github.com/api3dao/airkeeper/blob/main/src/call-api.ts) in Airkeeper can be
-re-used for this step.
+Probably current
+[call-api-ts](https://github.com/api3dao/airkeeper/blob/2de24dd82d710d1dc383049f2637a57b15d90ae2/src/call-api.ts) in
```
acenolaza, line 59: Since I've merge the changes from the refactor then we can keep pointing to main branch: 

```suggestion
[call-api-ts](https://github.com/api3dao/airkeeper/blob/main/src/api/call-api.ts) in
```

aquarat: Awesome thanks :)

psp_notes.md
```diff
@@ -55,7 +55,7 @@ https://github.com/api3dao/airnode/blob/v1-protocol/packages/airnode-node/src/ev
 
 5. The Airnode makes the API call specified by the template and the additional parameters
 
-Probably current [call-api-ts](https://github.com/api3dao/airkeeper/blob/main/src/call-api.ts) in Airkeeper can be
+Probably current [call-api-ts](https://github.com/api3dao/airkeeper/blob/main/src/api/call-api.ts) in Airkeeper can be
```
vponline, line 58:
```suggestion
Probably current [call-api.ts](https://github.com/api3dao/airkeeper/blob/main/src/api/call-api.ts) in Airkeeper can be
```
Should be `call-api.ts`?

acenolaza (approved):
🏆 Thanks for fixing this

psp_notes.md
aquarat: You're right, it's weird that the tests are now passing.

vponline: It can also be fixed from  #33  :smile:

aquarat: Sure, let's do that. I was going to leave it because they're psp_notes anyway, which will need to be processed into something suitable for public use.

PR #35, opened 2022-03-15
Retrygo refactor

vponline:
This replaces the `retryGo` function with `go` from [promise-utils](https://github.com/api3dao/promise-utils) and updates the code to use the new `GoResult` return value instead of the tuple form.

### DIFF
--- a/src/api/call-api.test.ts
+++ b/src/api/call-api.test.ts
@@ -136,9 +136,7 @@ describe('callApi', () => {
     const spy = jest.spyOn(adapter, 'buildAndExecuteRequest') as any;
 
     const apiResponse = { success: true, result: '723.392028' };
-    spy.mockResolvedValueOnce({
-      ...apiResponse,
-    });
+    spy.mockResolvedValueOnce(apiResponse);
 
     const [logs, res] = await callApi(callApiOptions);
 
@@ -245,9 +243,7 @@ describe('callApi', () => {
   it('returns an error if the API call fails to extract and encode response', async () => {
     const buildAndExecuteRequestSpy = jest.spyOn(adapter, 'buildAndExecuteRequest') as any;
     const apiResponse = { success: true, result: '723.392028' };
-    buildAndExecuteRequestSpy.mockResolvedValueOnce({
-      ...apiResponse,
-    });
+    buildAndExecuteRequestSpy.mockResolvedValueOnce(apiResponse);
 
     const extractAndEncodeResponseSpy = jest.spyOn(adapter, 'extractAndEncodeResponse');
     const error = new Error('Unexpected error');
--- a/src/api/call-api.test.ts
+++ b/src/api/call-api.test.ts
@@ -137,7 +137,7 @@ describe('callApi', () => {
 
     const apiResponse = { success: true, result: '723.392028' };
     spy.mockResolvedValueOnce({
-      data: apiResponse,
+      ...apiResponse,
     });
 
     const [logs, res] = await callApi(callApiOptions);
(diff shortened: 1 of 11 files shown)


src/api/call-api.test.ts
```diff
@@ -246,7 +246,7 @@ describe('callApi', () => {
     const buildAndExecuteRequestSpy = jest.spyOn(adapter, 'buildAndExecuteRequest') as any;
     const apiResponse = { success: true, result: '723.392028' };
     buildAndExecuteRequestSpy.mockResolvedValueOnce({
-      data: apiResponse,
+      ...apiResponse,
```
vponline, line 249: @acenolaza  Perhaps you could confirm if this `apiResponse` value structure change is fine (since [the `GoResult` and `GoResultSuccess`](https://github.com/api3dao/promise-utils/blob/main/src/index.ts#L4) formats are different now)

vponline: I think this is only changing the structure of the log message [here](https://github.com/api3dao/airkeeper/pull/35/files#diff-3e81ddd5a282280de51c1ef85c75c1275ed854660d5f853f13f8527ce6f0ff9fL63) because I am stringifying `apiResponse.data` instead of the whole `apiResponse` object (which has the structure of `{ success: true, data: {...} }`).

```ts
const message = `Failed to extract or encode value from API response: ${JSON.stringify(apiResponse.data)}`;
```

So stringifying `apiResponse` in the message would be another option if we want to include the `data: {...}` wrapper.

acenolaza: The actual `data` is now under `apiResponse.data.data` which is weird 🤔

acenolaza: > I think this is only changing the structure of the log message [here](https://github.com/api3dao/airkeeper/pull/35/files#diff-3e81ddd5a282280de51c1ef85c75c1275ed854660d5f853f13f8527ce6f0ff9fR65) because I am stringifying `apiResponse.data` instead of the whole `apiResponse` object (which has the structure of `{ success: true, data: {...} }`).
> 
> ```ts
> const message = `Failed to extract or encode value from API response: ${JSON.stringify(apiResponse.data)}`;
> ```
> 
> So stringifying `apiResponse` in the message would be another option if we want to include the `data: {...}` wrapper.

I don't think we want to JSON.stringify the whole AxiosResponse object

src/evm/initialize-provider.ts
```diff
@@ -35,14 +36,14 @@ export const initializeProvider = async (
...
+  if (!currentBlock.success) {
     const message = 'Failed to fetch the blockNumber';
-    const log = node.logger.pend('ERROR', message, errorGetBlockNumber);
+    const log = node.logger.pend('ERROR', message, currentBlock.error);
     return [[log], null];
   }
-  const currentBlockMessage = `Current block number for chainId ${chain.id}: ${currentBlock}`;
-  const currentBlockLog = node.logger.pend('INFO', currentBlockMessage);
+  const currentBlockMessage = `Current block number for chainId ${chain.id}: ${currentBlock.data}`;
+  const currentBlockLog = node.logger.pend('DEBUG', currentBlockMessage);
```
acenolaza, line 46: I was asked to make this INFO [here](https://github.com/api3dao/airkeeper/pull/31#discussion_r825887889). Same with gasTarget.

acenolaza (approved):
I didn't test this changes on my local env but implementation looks correct 👍🏻

src/evm/initialize-provider.ts
vponline: :+1:  Thanks for catching this... fixed

vponline:
I'll merge this now as the newer PRs will be changing these files significantly.

PR #36, opened 2022-03-15
Add all-tests-passed and docker image builds

aquarat:
It seems that having an "All tests passed" step is required to allow Github require CI status checks for PRs.
I added Docker builds while I was at it.


amarthadan (approved):
You always go the extra mile :joy:

PR #32, opened 2022-03-10
README polishing

aquarat:
I'm familiarising myself with the latest updates and while doing so I polished a few things.

PR #37, opened 2022-03-16
Refactor forEach to reduce

vponline:
Refactors `validSubscriptions` and `sponsorWalletsWithSubscriptions` to use `reduce`.
Updated to also refactor:  `enabledSubscriptions` and `groupedSubscriptions`.

~~Note: `enabledSubscriptions` and `groupedSubscriptions` are covered here: #32. I'm opening this so I can move on to other work but this PR can be closed if these are also to be covered in #32.~~

### DIFF
--- a/src/handlers/psp.ts
+++ b/src/handlers/psp.ts
@@ -244,7 +244,6 @@ const checkSubscriptionsConditions = async (
   voidSigner: ethers.VoidSigner,
   logOptions: node.LogOptions
 ) => {
-  const validSubscriptions: CheckedSubscription[] = [];
   const conditionPromises = subscriptions.map(
     (subscription) =>
       checkSubscriptionCondition(subscription, apiValuesBySubscriptionId[subscription.id], contract, voidSigner).then(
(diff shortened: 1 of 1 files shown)


src/handlers/psp.ts
```diff
@@ -289,30 +298,36 @@ const groupSubscriptionsBySponsorWallet = async (
...
-      sponsorWallet: data.sponsorWallet,
-    });
-  });
+      let nextNonce = data.transactionCount;
+      return [
+        ...acc,
+        {
+          subscriptions: subscriptionsBySponsor[data.sponsor].map((subscription) => ({
+            ...subscription,
+            nonce: nextNonce++,
```
aquarat, line 323: I don't know if this is an appropriate time to point this out or change this:
The `nextNonce` can reference the index of the `map`, like:
```
      return [
        ...acc,
        {
          subscriptions: subscriptionsBySponsor[data.sponsor].map((subscription, idx) => ({
            ...subscription,
            nonce: data.transactionCount + idx + 1,
          })),
```
this gets rid of the mutation.

aquarat (approved):
LGTM, see comment above if you're in the mood to change that too.

src/handlers/psp.ts
vponline: Seems reasonable :smile: ... fixed

acenolaza: I'm not sure this would work tbh because first transaction for a wallet should have a nonce of 0 and here we would always start at 1 🤔

vponline: @acenolaza Wow you're right, nice catch as always... ill fix

acenolaza: Fixed here: https://github.com/api3dao/airkeeper/commit/0b4957d993b3c50f8624039c4bddab208e8b97d6

I ❤️ the usage of reduce btw

PR #40, opened 2022-03-16
Fixes nonce assigment for subscriptions

acenolaza:
The index should be from the map and not reduce function. We should also start from 0 when assigning nonces.


acenolaza:
I don't think we need to revert the commit in [here](https://github.com/api3dao/airkeeper/pull/39) if we agree that this is the fix.

vponline (approved):
LGTM :+1:

PR #41, opened 2022-03-17
Fix derivation of subscriptionId

aquarat:
I discovered through a chat with Burak this morning that the subscription Id was being incorrectly derived, probably due to the Trail-of-Bits-related change (packing).

The related contract code is here: https://github.com/api3dao/airnode/blob/fb7968b0bd38a3ba85f6c2af5bae779a0b0df7a4/packages/airnode-protocol-v1/contracts/dapis/DapiServer.sol#L269


aquarat:
>  We should also update the script

oh yes! I will update that as part of this :heart:

PR #44, opened 2022-03-18
Airnode functions refactor

vponline:
This updates **wallet** functions, **loadConfig**, **logger**, **getGasPrice**, **callApi** to be imported from **airnode-node** v0.5

Note: to get this to build locally, you will need to copy the built **airnode-node** and **airnode-utilities** `/dist` folders from [Airnode PR 944](https://github.com/api3dao/airnode/pull/944) into your Airkeeper `node_modules/@api3`

This probably shouldn't be merged until we can update to v0.5


src/api/call-api.test.ts
```diff
@@ -1,291 +1,291 @@
-import * as abi from '@api3/airnode-abi';
-import * as adapter from '@api3/airnode-adapter';
-import * as node from '@api3/airnode-node';
-import * as ois from '@api3/airnode-ois';
-import { ethers } from 'ethers';
-import { callApi } from './call-api';
+// import * as abi from '@api3/airnode-abi';
```
vponline, line 1: @acenolaza  I've temporarily commented out these tests as `callApi` changed so much but I wanted to open this PR before I lose track of all the changes... sorry for it all being in one commit but I wasn't able to make it build otherwise :pray: 

I'll fix these tests next

acenolaza (changes requested):
I'm really looking forward to v0.5 packages being released 🚀

src/api/call-api.test.ts
acenolaza: No worries. I assumed that was the case. 👍🏻

src/handlers/psp.ts
```diff
@@ -166,17 +161,36 @@ const initializeState = (config: Config): State => {
...
+      endpointId: endpoint.id,
+      endpointName: endpoint.endpointName,
+      oisTitle: endpoint.oisTitle,
+      parameters: apiCallParameters,
+      // templateId: templateId,
+      // template: {
+      //   id: templateId,
+      //   ...template,
+      // },
+    };
```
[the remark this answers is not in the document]
vponline: :+1:  Reason I left them here was that I was not sure if we wanted to have the Airkeeper `callApi` wrapper function at all or go directly with` node.handlers.callApi`...

But thanks for confirming, I'll make the modifications :handshake:

src/handlers/psp.ts
```diff
@@ -166,17 +161,36 @@ const initializeState = (config: Config): State => {
...
   const { config, baseLogOptions, groupedSubscriptions } = state;
+  const airnodeHDNode = ethers.utils.HDNode.fromMnemonic(config.nodeSettings.airnodeWalletMnemonic);
+  const airnodeAddress = (
+    config.airnodeXpub
+      ? ethers.utils.HDNode.fromExtendedKey(config.airnodeXpub).derivePath('0/0')
+      : airnodeHDNode.derivePath(ethers.utils.defaultPath)
+  ).address;
+  if (config.airnodeAddress && config.airnodeAddress !== airnodeAddress) {
+    throw new Error(`xpub does not belong to Airnode: ${airnodeAddress}`);
+  }
```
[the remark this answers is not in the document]
vponline: > One last thing to notice is that node.callApi() only uses airnodeAddress for regular requests so it might even be unnecessary to even set this value.

I'll look into this as it may need some changes on the airnode side :thinking: 

> Another thing to consider is that subscription objects also have airnodeAddress as prop.
> Then we can just do const airnodeAddress = state.airnodeWallet.address

Oherwise I can do either of these

src/config.ts
```diff
@@ -8,21 +8,8 @@ import { AirkeeperConfig, validateConfig } from './validator';
...
-
-  if (shouldSkipValidation) {
-    return config;
-  }
-  if (!valid) {
-    throw new Error(`Invalid Airnode configuration file: ${JSON.stringify(messages, null, 2)}`);
-  }
-
-  return config;
+  //TODO: loadConfig can be wrapped in a custom validation function and logger can be used there to log possible errors thrown by node.config.loadConfig
```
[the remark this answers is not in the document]
vponline: Do you mean simply a try-catch block? Please see here for [loadConfig](https://github.com/api3dao/airnode/pull/944/files#diff-e3e940060e0ee2dfc73ef707a62ce607cf1a3a7459ce83c82d6621dbd674538fR26) or did you have another idea?

src/api/call-api.ts
```diff
@@ -1,67 +1,22 @@
...
-  const configEndpoint = configOis.endpoints.find((e) => e.name === endpointName)!;
-  const reservedParameters = node.adapters.http.parameters.getReservedParameters(configEndpoint, apiCallParameters);
-  if (!reservedParameters._type) {
-    const message = `reserved parameter '_type' is missing for endpoint: ${endpointName}`;
-    const log = node.logger.pend('ERROR', message);
-    return [[log], null];
+//TODO: check if Airkeeper's merged Config is ok or we should be using node.Config type
+export const callApi = async (payload: {
+  config: Config;
+  aggregatedApiCall: node.AggregatedApiCall;
```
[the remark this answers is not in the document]
vponline: Makes perfect sense :handshake:

src/api/call-api.ts
```diff
@@ -1,67 +1,22 @@
-import * as adapter from '@api3/airnode-adapter';
 import * as node from '@api3/airnode-node';
-import * as ois from '@api3/airnode-ois';
+import { logger } from '@api3/airnode-utilities';
```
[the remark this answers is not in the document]
vponline: No objections :smile:, I'll update

src/config.ts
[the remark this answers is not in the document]
acenolaza: I would prefer not to log things in this file but maybe we can try/catch and return `[logs, config]` or `[error, config]` something like that and the caller would do the logging. This is just an idea 😁

src/api/call-api.ts
```diff
@@ -1,67 +1,22 @@
...
-  const configOis = oises.find((o) => o.title === oisTitle)!;
-  const configEndpoint = configOis.endpoints.find((e) => e.name === endpointName)!;
-  const reservedParameters = node.adapters.http.parameters.getReservedParameters(configEndpoint, apiCallParameters);
-  if (!reservedParameters._type) {
-    const message = `reserved parameter '_type' is missing for endpoint: ${endpointName}`;
-    const log = node.logger.pend('ERROR', message);
-    return [[log], null];
+//TODO: check if Airkeeper's merged Config is ok or we should be using node.Config type
+export const callApi = async (payload: {
+  config: Config;
```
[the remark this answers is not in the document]
acenolaza: oh right! you are sending the whole payload object to node.callApi(). I assumed only aggregatedApiCall was required 🤦🏻‍♂️

src/api/call-api.test.ts
```diff
@@ -188,58 +65,54 @@ describe('callApi', () => {
...
       expect.arrayContaining([
         {
           level: 'ERROR',
-          message: expect.stringMatching("reserved parameter '_type' is missing for endpoint: convertToUSD"),
+          message: "Cannot read property 'length' of undefined",
+        },
+        {
+          error: "Cannot read property 'length' of undefined",
+          level: 'ERROR',
+          message:
```
vponline, line 82: So due to the double logs from `callApi` these are still incorrect, 

Should we just go with the default (error) logs returned by` node.handlers.callApi`, or use the more detailed logs @acenolaza you had before. Mainly, we no longer have access to the `apiResponse` values :thinking:

acenolaza: This is now testing node.callApi() internals so maybe that should be tested in the node package but I'm surprised to see that this check has been removed from the [buildOptions()](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/api/index.ts#L13) function.

vponline: hmm what kind of testing should we do here instead? just `callApi` with mocked success and error cases?

acenolaza: Right, this function is now much simpler so I guess most tests we used to have might no longer apply

src/handlers/psp.ts
```diff
@@ -25,24 +27,32 @@ import {
...
 
 export const handler = async (_event: any = {}): Promise<any> => {
   const startedAt = new Date();
 
-  const airnodeConfig = loadAirnodeConfig();
+  const airnodeConfig = goSync(loadAirnodeConfig);
+  if (!airnodeConfig.success) {
+    utils.logger.error(airnodeConfig.error.message);
+    throw airnodeConfig.error;
+  }
```
vponline, line 41: > I would prefer not to log things in this file but maybe we can try/catch and return [logs, config] or [error, config] something like that and the caller would do the logging. This is just an idea

Since we don't want to log anything in **config.ts**, how about using the `goSync` function from **promise-utils** as above?

This way we don't need to modify the loadConfig functions to return tuples. :thinking: 

What do you think? @acenolaza

src/handlers/psp.ts
```diff
@@ -166,48 +176,51 @@ const initializeState = (config: Config): State => {
...
-            apiValue: ethers.BigNumber | null;
-            subscriptions: Id<Subscription>[];
-          }>
-      )
+
+    return go(
+      () =>
+        callApi(config, {
+          id: template.id,
+          airnodeAddress: subscriptions.find((s) => s.templateId === Object.keys(template)[0])!.airnodeAddress,
```
vponline, line 186: @acenolaza  Would this work for the PSP airnodeAddress? I think otherwise we would need to change the types in Airnode.

src/api/call-api.ts
```diff
@@ -1,67 +1,28 @@
-import * as adapter from '@api3/airnode-adapter';
```
amarthadan, line 1: Do we even still need this file? I was hoping we would get rid of it once we use the `airnode-node` version.

src/handlers/psp.ts
```diff
@@ -166,48 +176,51 @@ const initializeState = (config: Config): State => {
...
+          parameters: apiCallParameters,
+        }).then(
+          ([logs, data]) =>
+            [logs, { templateId: template.id, apiValue: data, subscriptions }] as node.LogsData<{
+              templateId: string;
+              apiValue: ethers.BigNumber | null;
+              subscriptions: Id<Subscription>[];
+            }>
+        ),
+      { timeoutMs: DEFAULT_RETRY_TIMEOUT_MS }
```
amarthadan, line 199: Since this is replacing `retryGo` I think it should actually retry the call, shouldn't it :smile: Maybe we can also use a better naming for the timeout constant.

src/handlers/rrp.ts
```diff
@@ -24,90 +25,102 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+        return callApi(config, {
+          id: templateId,
+          airnodeAddress,
+          endpointId,
+          endpointName,
+          oisTitle,
+          parameters: apiCallParameters,
+        }).then(([logs, data]) => [logs, { [beaconId]: data }] as node.LogsData<ApiValueByBeaconId>);
+      },
+      { timeoutMs: DEFAULT_RETRY_TIMEOUT_MS }
```
amarthadan, line 106: Same here.

src/handlers/rrp.ts
```diff
@@ -148,19 +161,21 @@ export const handler = async (_event: any = {}): Promise<any> => {
         const rrpBeaconServer = protocol.RrpBeaconServerFactory.connect(chain.contracts.RrpBeaconServer!, provider);
 
         // Fetch current block number from chain via provider
-        const [err, currentBlock] = await retryGo(() => provider.getBlockNumber());
-        if (err || isNil(currentBlock)) {
-          node.logger.error('failed to fetch the blockNumber', {
+        const currentBlock = await go(() => provider.getBlockNumber(), {
+          timeoutMs: DEFAULT_RETRY_TIMEOUT_MS,
```
amarthadan, line 165: Same here.

src/handlers/rrp.ts
```diff
@@ -188,24 +201,25 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
           // 3.2.2 Fetch keeperSponsorWallet transaction count
           // **************************************************************************
-          node.logger.debug('fetching transaction count...', keeperSponsorWalletLogOptions);
+          utils.logger.debug('fetching transaction count...', keeperSponsorWalletLogOptions);
 
-          const [err, keeperSponsorWalletTransactionCount] = await retryGo(() =>
-            provider.getTransactionCount(keeperSponsorWallet.address, currentBlock)
+          const keeperSponsorWalletTransactionCount = await go(
+            () => provider.getTransactionCount(keeperSponsorWallet.address, currentBlock.data),
+            { timeoutMs: DEFAULT_RETRY_TIMEOUT_MS }
```
amarthadan, line 208: Same here.

src/handlers/rrp.ts
```diff
@@ -274,48 +288,48 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
 
             // address(0) is considered whitelisted
             const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
-            const [errReadBeacon, beaconResponse] = await retryGo(() =>
-              rrpBeaconServer.connect(voidSigner).readBeacon(beaconId)
-            );
-            if (errReadBeacon || isNil(beaconResponse) || isNil(beaconResponse.value)) {
-              node.logger.error(`failed to read value for beaconId: ${beaconId}`, {
+            const beaconResponse = await go(() => rrpBeaconServer.connect(voidSigner).readBeacon(beaconId), {
+              timeoutMs: DEFAULT_RETRY_TIMEOUT_MS,
```
amarthadan, line 296: Same here.

src/handlers/rrp.ts
```diff
@@ -337,73 +351,76 @@ export const handler = async (_event: any = {}): Promise<any> => {
               requestSponsor,
               keeperSponsorWallet.address
             );
-            const [errRequestedBeaconUpdateFilter, requestedBeaconUpdateEvents] = await retryGo(() =>
-              rrpBeaconServer.queryFilter(requestedBeaconUpdateFilter, blockHistoryLimit * -1, currentBlock)
+            const requestedBeaconUpdateEvents = await go(
+              () => rrpBeaconServer.queryFilter(requestedBeaconUpdateFilter, blockHistoryLimit * -1, currentBlock.data),
+              { timeoutMs: DEFAULT_RETRY_TIMEOUT_MS }
```
amarthadan, line 356: Same here.

src/handlers/rrp.ts
```diff
@@ -337,73 +351,76 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
               continue;
             }
 
             // Fetch UpdatedBeacon events by beaconId
             const updatedBeaconFilter = rrpBeaconServer.filters.UpdatedBeacon(beaconId);
-            const [errUpdatedBeaconFilter, updatedBeaconEvents] = await retryGo(() =>
-              rrpBeaconServer.queryFilter(updatedBeaconFilter, blockHistoryLimit * -1, currentBlock)
+            const updatedBeaconEvents = await go(
+              () => rrpBeaconServer.queryFilter(updatedBeaconFilter, blockHistoryLimit * -1, currentBlock.data),
+              { timeoutMs: DEFAULT_RETRY_TIMEOUT_MS }
```
amarthadan, line 370: Same here.

src/handlers/rrp.ts
```diff
@@ -337,73 +351,76 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+            const [pendingRequestedBeaconUpdateEvent] = requestedBeaconUpdateEvents.data.filter(
+              (rbue) => !updatedBeaconEvents.data.some((ub) => rbue.args!['requestId'] === ub.args!['requestId'])
             );
             if (!isNil(pendingRequestedBeaconUpdateEvent)) {
               // Check if RequestedBeaconUpdate event is awaiting fulfillment by calling AirnodeRrp.requestIsAwaitingFulfillment with requestId and check if beacon value is fresh enough and skip if it is
-              const [errRequestIsAwaitingFulfillment, requestIsAwaitingFulfillment] = await retryGo(() =>
-                airnodeRrp.requestIsAwaitingFulfillment(pendingRequestedBeaconUpdateEvent.args!['requestId'])
+              const requestIsAwaitingFulfillment = await go(
+                () => airnodeRrp.requestIsAwaitingFulfillment(pendingRequestedBeaconUpdateEvent.args!['requestId']),
+                { timeoutMs: DEFAULT_RETRY_TIMEOUT_MS }
```
amarthadan, line 388: Same here.

src/handlers/rrp.ts
```diff
@@ -417,28 +434,30 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+                rrpBeaconServer
+                  .connect(keeperSponsorWallet)
+                  .requestBeaconUpdate(
+                    templateId,
+                    requestSponsor,
+                    requestSponsorWallet.address,
+                    encodedParameters,
+                    overrides
+                  ),
+              { timeoutMs: DEFAULT_RETRY_TIMEOUT_MS }
```
amarthadan, line 448: Same here.

amarthadan (changes requested):

src/api/call-api.ts
vponline: The idea was to use this as a centralized place to construct the node.handlers.callApi inputs instead of doing these inline in the handlers since they work a bit differently as well as parse the result value (see [here](https://github.com/api3dao/airkeeper/pull/44#discussion_r829983119))

Please let me know if you'd prefer to change this

amarthadan: Alright, makes sense :slightly_smiling_face:

src/handlers/psp.ts
```diff
@@ -165,45 +176,63 @@ const initializeState = (config: Config): State => {
...
+
+    return go(
       () =>
-        callApi({
-          oises: config.ois,
-          apiCredentials: config.apiCredentials,
-          apiCallParameters,
-          oisTitle: endpoint.oisTitle,
+        callApi(config, {
+          id: template.id,
```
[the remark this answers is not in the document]
vponline: > I'm starting to really dislike the node.CallApiPayload and how tied it is to RRPrequests.
> This should work but I find it weird that we need to pass the airnodeAddress here since it's only required for '[regular](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/api/index.ts#L36)' request type.

> Now that I see this I'm remembering the verification steps in node.callApi()
1. I think it doesn't make sense to send templateId as requestId
2. verification of requestId won't make sense
3. verification of the sponsor wallet seems that will be done twice if we call node.callApi()

I think the above points are both related to the types required by node.callApi() so we could simplify `callApiOptions` if we were to make these fields optional in the **airnode-node** `AggregatedApiCall` type.

@acenolaza @amarthadan Any objections for me to make these (type) changes to Airnode? Or please let me know if you have other ideas.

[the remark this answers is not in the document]
amarthadan: AFAIK no verification will be done for `http-gateway` type we want to use here.

I agree that the current design of `callApi` is not really prepared for a wide variety of requests and that should be changed. But for now I would suggest keeping up with the current `callApi` interface even though some values might not make sense. I'll create a task for future refactoring of `callApi`. Do you agree?

package.json
```diff
@@ -3,13 +3,17 @@
...
+    "node": "^14.17.1"
+  },
   "main": "./dist/src/start.js",
   "files": [
     "dist",
     "serverless.yml",
     "*.js"
   ],
   "scripts": {
+    "postinstall": "git clone https://github.com/api3dao/airnode.git airnode; cd airnode && yarn run bootstrap && yarn build && cd ..; cp -a airnode/node_modules/@api3/airnode-node/. node_modules/@api3/airnode-node; rm -rf airnode",
```
amarthadan, line 16: I'm not sure about this tbh. If there's no simple way to do this I wouldn't force it.
Since we all have the airnode cloned anyway what about this. We'll remove the postinstall so nothing is done automatically but we add `link:airnode` target that would expect a path (both relative and absolute) to the local `airnode` directory and links the `airnode/node_modules/@api3` to `node_modules/@api3`? It would require a tiny node script to do that but that's like 5 lines of code. WDYT?

vponline: I can make these changes, but this wouldn't work for github tests, right? :thinking:

vponline (commented):
@acenolaza Do the example configs work for you in the main branch? I had to change the `airnodeAddress` in **airkeeper.json** ([these changes](https://github.com/api3dao/airkeeper/pull/44/commits/58fbfa6fa0a5650980784f58f2cba4aa7867cdc3)) to make them work.

package.json
amarthadan: :thinking: You're right, that's true. Ok, let's go with your approach, it's temporary anyway.

vponline: or alternatively request a 0.5.1 release for airnode-node? :smile:

amarthadan: Unfortunately, I think this is not the last time we're having this problem while we don't have something at least semi-stable. We'll see.

src/handlers/psp.ts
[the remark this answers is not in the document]
acenolaza: You are right @amarthadan about the verification being skipped when using http-gateway. I missed the checks that return null on each verify* function. i. e. https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/api/index.ts#L130

I'm on board with the idea to make it work and then we make it pretty but I would prefer if we keep the [callApi](https://github.com/api3dao/airkeeper/blob/main/src/api/call-api.ts#L9)() interface clean and simple so rrp.ts and psp.ts would call this function the same way and with just the minimum required args. Then this required by the node.callApi() can be hard-coded/added inside this function and if we then change the node side no changes would be required in rrp.ts and psp.ts...just in call-api.ts. This is only a suggestion and can be disregarded at will.

acenolaza:
> @acenolaza Do the example configs work for you in the main branch? I had to change the `airnodeAddress` in **airkeeper.json** ([these changes](https://github.com/api3dao/airkeeper/pull/44/commits/58fbfa6fa0a5650980784f58f2cba4aa7867cdc3)) to make them work.

I'm actually using 0xA30CA71Ba54E83127214D3271aEA8F5D6bD4Dace because in secrets.env I have the test mnemonic we are using everywhere else

src/handlers/psp.ts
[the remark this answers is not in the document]
amarthadan: Totally agreed, let's do it that way :muscle:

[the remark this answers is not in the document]
vponline: > I'm on board with the idea to make it work and then we make it pretty but I would prefer if we keep the [callApi](https://github.com/api3dao/airkeeper/blob/main/src/api/call-api.ts#L9)() interface clean and simple so rrp.ts and psp.ts would call this function the same way and with just the minimum required args.

Sounds great :smile:

package.json
vponline: That's a great point and makes sense to add

scripts/psp-beacon-local-setup.js
```diff
@@ -57,9 +57,9 @@ async function main() {
   // Wallets
   const airnodeWallet = ethers.Wallet.fromMnemonic(config.airnodeMnemonic);
   console.log('👛 ~ airnodeWallet', airnodeWallet.address);
-  const airnodePspSponsorWallet = deriveSponsorWallet(config.airnodeMnemonic, roles.sponsor.address, '2').connect(
-    provider
-  );
+  const airnodePspSponsorWallet = evm
+    .deriveSponsorWalletFromMnemonic(config.airnodeMnemonic, roles.sponsor.address, '2')
```
amarthadan, line 61: I think you can use a constant from `constants.ts` for the protocol index.

src/evm/check-conditions.ts
```diff
@@ -43,20 +44,27 @@ export const checkSubscriptionCondition = async (
         .connect(voidSigner)
         .functions[conditionFunction.name](subscription.id, encodedFulfillmentData, conditionParameters),
     {
-      timeoutMs: DEFAULT_RETRY_TIMEOUT_MS,
+      timeoutMs: TIMEOUT_MS,
+      retries: 1,
```
amarthadan, line 48: This can be also a constant in `constants.ts`.
I'm not sure how did the previous `retryGo` implementation look like, was it just one retry?

amarthadan (approved):
LGTM, just a bit of tiny polishing :slightly_smiling_face:

src/evm/check-conditions.ts
vponline: The previous **retryGo** was:

```ts
export const retryGo = <T>(fn: () => Promise<T>, options?: node.utils.PromiseOptions) =>
  node.utils.go(() => node.utils.retryOnTimeout(DEFAULT_RETRY_TIMEOUT_MS, fn), options);
```
and **retryOnTimeout** was:

```ts
export function retryOnTimeout<T>(maxTimeoutMs: number, operation: () => Promise<T>, options?: ContinuousRetryOptions) {
  const promise = new Promise<T>((resolve, reject) => {
    function run(): Promise<any> {
      // If the promise is successful, resolve it and bubble the result up
      return operation()
        .then(resolve)
        .catch((reason: any) => {
          // Only if the error is a timeout error, do we retry the promise
          if (reason instanceof Error && reason.message.includes('Operation timed out')) {
            // Delay the new attempt slightly
            return sleep(options?.delay || DEFAULT_RETRY_DELAY_MS)
              .then(run)
              .then(resolve)
              .catch(reject);
          }

          // If the error is NOT a timeout error, then we reject immediately
          return reject(reason);
        });
    }

    return run();
  });

  return promiseTimeout(maxTimeoutMs, promise);
}
```

src/evm/process-sponsor-wallet.ts
```diff
@@ -62,7 +62,7 @@ export const processSponsorWallet = async (
               nonce,
             }
           ),
-      { timeoutMs: TIMEOUT_MS, retries: RETRIES }
```
vponline: @amarthadan @acenolaza  I don't understand why adding retries here is causing `transactions-count.test.ts` to fail :confused:

acenolaza: I think the problem is that I used the `mockRejectedValueOnce` function and we are now callling it more than once causing the test to timeout. You could try using `mockRejectedValue` or `mockImplementation`

PR #45, opened 2022-03-18
Fixes contract abis and adds LogOptions factory method

acenolaza:
Sorry in advance for putting everything in the same PR but they seem to be small enough 😁 

1. Updates contract ABI in order to make PSP work again. (I forgot to tell you about that @aquarat when you fixed the [encoding issue](https://github.com/api3dao/airkeeper/commit/e9f4f7e57917c80523521ac6fb56323037c8c911))
2. Adds a method to utils that can be used to create LogOptions objects with overrides. (I'm hoping this will make you happy @aquarat )
3. Better gas target log message


aquarat:
I'm reviewing this now :)

src/evm/initialize-provider.ts
```diff
@@ -55,7 +54,15 @@ export const initializeProvider = async (
     const log = node.logger.pend('ERROR', message);
     return [[log], null];
   }
-  const gasTargetMessage = `Gas target for chainId ${chain.id}: ${JSON.stringify(gasTarget)}`;
+  let gasTargetMessage;
```
aquarat, line 57: There's an opportunity to make this a separate function - but I have a feeling this will be replaced with an Airnode dependency, so probably unimportant.

aquarat (approved):
Greatly improved :) 

For a future date perhaps you (or we) could use a factory function where it returns an object with a set of logging functions (like warn, error, etc) and these have `baseLogOptions` pre-loaded so-to-speak, so that you don't have to pass around `baseLogOptions` and you won't have to use `buildLogOptions`. You'd call `logger.warn('Failed to fetch API value. Skipping update...', {additional: { templateId: data.templateId }})` as an example perhaps. I think this is a pending issue in Airnode too, I'm not sure what the approach will be though.

But there's limited business value in this currently. It's just an idea for the future :)

acenolaza:
> 

That's a really interesting idea 🤔 Do you see this being implemented on the node side where logger lives or are you thinking about wrapping the logger in Airkeeper to add this functionality?

aquarat:
> > 
> 
> That's a really interesting idea thinking Do you see this being implemented on the node side where logger lives or are you thinking about wrapping the logger in Airkeeper to add this functionality?

I think this might be something implemented in Airnode - I know there's an open task along the lines of "Make it so that logger config doesn't have to be passed around" in Airnode, so I suspect this is the same underlying issue. :)

src/logger.ts
```diff
@@ -0,0 +1,28 @@
+import * as node from '@api3/airnode-node';
+
+type LogOptionsOverride = 'meta' | 'additional';
+
+export const buildLogOptions = (
+  override: LogOptionsOverride,
+  fields: { [key: string]: unknown },
+  baseLogOptions: node.LogOptions
+): node.LogOptions => {
```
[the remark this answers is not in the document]
acenolaza: Right but I think it provides better typing on the caller. When I hover over the variable being assigned the result of calling this function it shows a dynamic type with all the props instead of node.LogOptions if I remove it. If you still feel like it is redundant the I'll remove it anyways.

src/handlers/psp.ts
```diff
@@ -166,44 +165,53 @@ const initializeState = (config: Config): State => {
...
     );
-    if (!logsData.success) {
-      node.logger.warn('Failed to fecth API value', templateLogOptions);
-      continue;
+  });
+  const responses = await Promise.all(apiValuePromises);
+
+  const apiValuesBySubscriptionId = responses.reduce((acc: { [subscriptionId: string]: ethers.BigNumber }, result) => {
+    if (!result.success) {
+      node.logger.warn('Failed to fecth API value', baseLogOptions);
```
aquarat, line 193: Oh yes, almost forgot about this spelling mistake @ fetch.

src/evm/initialize-provider.ts
acenolaza: I basically copy pasted that code snipped from [airnode-node](https://github.com/api3dao/airnode/blob/master/packages/airnode-node/src/evm/handlers/process-transactions.ts#L45-L55) but you make a good point 🤔 

Since getGasPrice() is already returning an array of logs (return type is LogsData) then maybe it would be a good idea to include that snipped in the logs returned by this function instead of doing that formatting on the caller side. WDYT?

PR #47, opened 2022-03-21
Fixes callApi response object and updates contracts ABI

acenolaza:
main branch is currently broken and changes in this PR fixes it

There's no jira issue for this but I discovered while working on BEC-229


vponline (approved):
Thanks for fixing this 🙏

PR #50, opened 2022-03-25
Refactor submitTransactions to process subscriptions sequentially

vponline:
This moves the subscription checking, wallet functions and processing to be done for each sponsorAddress group.
Opening this as draft to get some help on how to proceed.

The old flow was:
1. Get all subscriptions from **airkeeper.json**
2. Filter subscriptions by chain ID (i.e. chain provider)
3. Check that all filtered subscriptions pass the condition check
4. Group subscriptions by sponsor wallet address
5. Start fulfilling subscriptions for each sponsor wallet-group in parallel

The new intended flow:
1. Get all subscriptions from **airkeeper.json**
2. Filter subscriptions by chain provider
3. Group subscriptions by **sponsorAddress**
4. Start processing the subscriptions for each **sponsorAddress** sequentially
   4.1. `getSponsorWalletAndTransactionCount` 
   4.2. if valid wallet data `processSponsorWallet`
      4.2.1. `checkSubscriptionCondition`
      4.2.2. if valid subscription `fulfillPspBeaconUpdate`
5. Process subscriptions for each provider

### DIFF
--- a/src/evm/process-sponsor-wallet.ts
+++ b/src/evm/process-sponsor-wallet.ts
@@ -2,6 +2,7 @@ import * as node from '@api3/airnode-node';
 import * as utils from '@api3/airnode-utilities';
 import { go } from '@api3/promise-utils';
 import { ethers } from 'ethers';
+import { checkSubscriptionCondition } from './check-conditions';
 import { GAS_LIMIT, TIMEOUT_MS, RETRIES } from '../constants';
 import { ProcessableSubscription } from '../types';
 
@@ -10,13 +11,33 @@ export const processSponsorWallet = async (
   contract: ethers.Contract,
   gasTarget: node.GasTarget,
   subscriptions: ProcessableSubscription[],
-  sponsorWallet: ethers.Wallet
+  sponsorWallet: ethers.Wallet,
+  voidSigner: ethers.VoidSigner,
+  apiValuesBySubscriptionId: { [subscriptionId: string]: ethers.BigNumber }
 ): Promise<node.LogsData<ProcessableSubscription>[]> => {
   const logs: node.LogsData<ProcessableSubscription>[] = [];
+  const sortedSubscriptions = subscriptions.sort((a, b) => a.nonce - b.nonce);
 
+  // Keep track of nonce outside of the loop in case there is an invalid subscription and its nonce is skipped
+  //TODO: improve nonce?
+  let nextNonce = sortedSubscriptions[0].nonce;
   // Process each subscription in serial to keep nonces in order
-  for (const subscription of subscriptions.sort((a, b) => a.nonce - b.nonce)) {
-    const { id: subscriptionId, relayer, sponsor, fulfillFunctionId, nonce, apiValue } = subscription;
+  for (const subscription of sortedSubscriptions) {
+    const { id: subscriptionId, relayer, sponsor, fulfillFunctionId, nonce } = subscription;
+    const apiValue = apiValuesBySubscriptionId[subscription.id];
+
+    // Check subscription
+    const [checkSubscriptionLogs, isValid] = await checkSubscriptionCondition(
+      subscription,
+      apiValue,
+      contract,
+      voidSigner
+    );
+    logs.push([checkSubscriptionLogs, subscription]);
+    // Skip processing if if subscription is invalid
+    if (!isValid) {
+      continue;
+    }
 
     // Encode API value
     const encodedFulfillmentData = ethers.utils.defaultAbiCoder.encode(['int256'], [apiValue]);
@@ -42,7 +63,8 @@ export const processSponsorWallet = async (
     } catch (error) {
       const message = 'Failed to get fulfill function';
       const log = utils.logger.pend('ERROR', message, error as any);
-      return [...logs, [[log], subscription]];
+      logs.push([[log], subscription]);
+      continue;
     }
     const tx = await go<ethers.ContractTransaction, Error>(
       () =>
@@ -59,7 +81,7 @@ export const processSponsorWallet = async (
             {
               gasLimit: GAS_LIMIT,
               ...gasTarget,
-              nonce,
+              nonce: nextNonce++,
             }
           ),
       { timeoutMs: TIMEOUT_MS, retries: RETRIES }
@@ -67,7 +89,8 @@ export const processSponsorWallet = async (
     if (!tx.success) {
       const message = `Failed to submit transaction using wallet ${sponsorWallet.address} with nonce ${nonce}`;
       const log = utils.logger.pend('ERROR', message, tx.error);
-      return [...logs, [[log], subscription]];
+      logs.push([[log], subscription]);
+      continue;
     }
     const message = `Tx submitted: ${tx.data.hash}`;
     const log = utils.logger.pend('INFO', message);
--- a/src/evm/process-sponsor-wallet.ts
+++ b/src/evm/process-sponsor-wallet.ts
@@ -12,8 +12,7 @@ export const processSponsorWallet = async (
   gasTarget: node.GasTarget,
   subscriptions: ProcessableSubscription[],
   sponsorWallet: ethers.Wallet,
-  voidSigner: ethers.VoidSigner,
-  apiValuesBySubscriptionId: { [subscriptionId: string]: ethers.BigNumber }
+  voidSigner: ethers.VoidSigner
 ): Promise<node.LogsData<ProcessableSubscription>[]> => {
   const logs: node.LogsData<ProcessableSubscription>[] = [];
   const sortedSubscriptions = subscriptions.sort((a, b) => a.nonce - b.nonce);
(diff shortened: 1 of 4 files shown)


src/evm/process-sponsor-wallet.ts
```diff
@@ -10,13 +11,33 @@ export const processSponsorWallet = async (
...
+      subscription,
+      apiValue,
+      contract,
+      voidSigner
+    );
+    logs.push([checkSubscriptionLogs, subscription]);
+    // Skip processing if if subscription is invalid
+    if (!isValid) {
+      continue;
+    }
```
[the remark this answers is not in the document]
vponline: This could probably be simplified with a `.reduce()` as well :thinking:

vponline:
So if I've understood correctly, the next step is to combine steps 5. and 6. (Promise.all(sponsorWalletPromises) and Promise.all(providerPromises)) into one array, and then invoke new Lambdas for each element in that array?

src/evm/process-sponsor-wallet.ts
```diff
@@ -10,13 +11,31 @@ export const processSponsorWallet = async (
...
-  sponsorWallet: ethers.Wallet
+  sponsorWallet: ethers.Wallet,
+  voidSigner: ethers.VoidSigner
 ): Promise<node.LogsData<ProcessableSubscription>[]> => {
   const logs: node.LogsData<ProcessableSubscription>[] = [];
+  const sortedSubscriptions = subscriptions.sort((a, b) => a.nonce - b.nonce);
 
+  // Keep track of nonce outside of the loop in case there is an invalid subscription and its nonce is skipped
+  //TODO: improve nonce?
+  let nextNonce = sortedSubscriptions[0].nonce;
```
amarthadan, line 22: I like that the nonces are not bound to the subscriptions and so we can skip invalid once. Let's see if there's some issue with that. But in case it works, we should maybe rename the field to something like `index` or `order` because the nonce doesn't have to be the nonce actually used which is misleading. Also, the (potentially) wrong nonce is logged.

src/handlers/psp.ts
```diff
@@ -256,156 +248,117 @@ const initializeProviders = async (state: State): Promise<State> => {
...
 };
 
-const checkSubscriptionsConditions = async (
-  subscriptions: Id<Subscription>[],
-  apiValuesBySubscriptionId: { [subscriptionId: string]: ethers.BigNumber },
-  contract: ethers.Contract,
-  voidSigner: ethers.VoidSigner,
-  logOptions: utils.LogOptions
+const processSubscriptions = async (
+  providerSubscriptions: {
```
amarthadan, line 252: I think you can create a separate type for this.

src/handlers/psp.ts
```diff
@@ -256,156 +248,117 @@ const initializeProviders = async (state: State): Promise<State> => {
...
-        );
-        utils.logger.logPending(logs, subscriptionLogOptions);
-      });
-    });
+      // Group filtered subscriptions by sponsorAddress
+      const subscriptionsBySponsor = groupBy(chainSubscriptions, 'sponsor');
 
-    await Promise.all(sponsorWalletPromises);
-  });
+      // Collect subscriptions for each provider
```
amarthadan, line 348:
```suggestion
      // Collect subscriptions for each provider + sponsor pair
```

src/handlers/psp.ts
```diff
@@ -256,156 +248,117 @@ const initializeProviders = async (state: State): Promise<State> => {
...
-        contracts['DapiServer'],
-        gasTarget,
-        subscriptions,
-        sponsorWallet
+const submitTransactions = async (state: State) => {
+  const { baseLogOptions, groupedSubscriptions, apiValuesBySubscriptionId, providerStates } = state;
+
+  const subscriptions = groupedSubscriptions.flatMap((s) => s.subscriptions);
+
+  const providerSubscriptions = providerStates.reduce(
```
amarthadan, line 325:
```suggestion
  const providerSponsorSubscriptions = providerStates.reduce(
```
Calling it just `providerSubscriptions` sounds misleading to me because you can have multiple objects with the same provider but different sponsor. That's actually the whole point, grouping it by provider + sponsor pair.

src/handlers/psp.ts
```diff
@@ -256,156 +248,117 @@ const initializeProviders = async (state: State): Promise<State> => {
...
-  );
-
-  return sponsorWalletsWithSubscriptions;
-};
-
-const submitTransactions = async (state: State) => {
-  const { baseLogOptions, groupedSubscriptions, apiValuesBySubscriptionId, providerStates } = state;
-
-  const providerPromises = providerStates.map(async (providerState) => {
+  const providerPromises = providerSubscriptions.map(async (subscriptionGroup) => {
```
amarthadan, line 259:
```suggestion
  const providerSponsorPromises = providerSubscriptions.map(async (subscriptionGroup) => {
```

src/handlers/psp.ts
amarthadan, line 252:
```suggestion
  providerSponsorSubscriptions: {
```

amarthadan (changes requested):
I think the code is fine but I was a bit confused by your PR description. The subscriptions are grouped by the provider + sponsor pair and processed in these groupings, right? At least that's how I understood the code.

Anyway, if either @acenolaza or @aquarat can take a look as well that would be great.

amarthadan:
> So if I've understood correctly, the next step is to combine `Promise.all(sponsorWalletPromises)` and `Promise.all(providerPromises)` into one array, and then invoke new Lambdas for each element in that array?

I don't really understand. Where did you get `Promise.all(sponsorWalletPromises)`?
Also, you can't really pass a promise to Lambda function, it would have to be data needed for processing. But yes, the concept is the same.

These are the promises that would need to be turned into Lambda invocations
https://github.com/api3dao/airkeeper/blob/222d3ace95a743d2e081072ef21098898761d653/src/handlers/psp.ts#L317

src/evm/process-sponsor-wallet.ts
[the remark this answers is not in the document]
vponline: :+1:  Ok will leave it as is (I was just thinking that a reduce could return the `logs`-array)

vponline:
>  it would have to be data needed for processing. But yes, the concept is the same.
These are the promises that would need to be turned into Lambda invocations

Sorry that was poorly worded, but this still answered the question I was going for  :smile:

src/evm/process-sponsor-wallet.ts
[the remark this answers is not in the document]
amarthadan: Oh, right, I haven't noticed the logs. But I still prefer this way in this case. But I'll leave that up to you, I don't have a strong preference :slightly_smiling_face:

src/handlers/psp.ts
```diff
@@ -256,156 +249,113 @@ const initializeProviders = async (state: State): Promise<State> => {
...
-      contracts['DapiServer'],
-      voidSigner,
+    // Skip processing for the current sponsorAddress if the wallet functions fail
+    if (isNil(walletData)) {
+      const sponsorLogOptions = buildLogOptions('additional', { sponsor: sponsorAddress }, providerLogOptions);
+      utils.logger.warn('Failed to fetch sponsor wallet or transaction count', sponsorLogOptions);
+      return;
+    }
+
+    const sponsorWallet = walletData.sponsorWallet;
```
aquarat, line 278: It's super minor but this could be `const { sponsorWallet } = walletData;`, feel free to leave it for next time.

aquarat (approved):
LGTM :rocket: 
Very neat work :)

acenolaza (commented):
Just wondering, have you tested this locally?

src/handlers/psp.ts
```diff
@@ -256,156 +249,113 @@ const initializeProviders = async (state: State): Promise<State> => {
...
     );
+    utils.logger.logPending(transactionCountLogs, sponsorWalletLogOptions);
 
-    // Group subscriptions by sponsor address
-    const subscriptionsBySponsor = groupBy(validSubscriptions, 'sponsor');
+    // Get subscriptions for the current sponsorAddress and assign nonces
+    const subscriptionsWithNonce = subscriptions.map((subscription, idx) => ({
+      ...subscription,
+      nonce: walletData.transactionCount + idx,
+    }));
```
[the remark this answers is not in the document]
vponline: > Just wondering, have you tested this locally?
Yes using the local-PSP, although that only has 1 subscription.

> I know that in processSponsorWallet() we take the initial nonce from the zero-index sorted array of subscriptions. What I'm thinking is that we might just want to send the sponsor wallet tx count as arg to processSponsorWallet() and avoid assigning nonces alltogether

:smile:  You're absolutely right and your suggestion is much nicer, I'll update.

src/evm/process-sponsor-wallet.ts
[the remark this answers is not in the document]
vponline: Let's leave as is then :smile:

amarthadan (approved):
LGTM :+1:

PR #53, opened 2022-03-29
Spawn new lambdas for each `providerSponsorSubscription`

vponline:
This updates PSP `updateBeacon` to spawn new lambdas to process each `providerSponsorSubscription` group.


.eslintrc.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
     'eslint:recommended',
     'plugin:@typescript-eslint/recommended',
     'plugin:@typescript-eslint/eslint-recommended',
-    'plugin:import/recommended',
+    // 'plugin:import/recommended', // This is causing a problem with importing types from aws-lambda package
```
vponline, line 20: I haven't spent much time looking into what the problem is, but VScode says it cannot resolve the module even though it is installed.

Do we need this setting?

src/handlers/psp.ts
```diff
@@ -249,68 +250,48 @@ const initializeProviders = async (state: State): Promise<State> => {
...
+        lambda.invoke(options, (err, data) => {
+          console.log('spawn', err, data);
+          if (err) {
+            reject(err);
+            return;
+          }
+          resolve(JSON.parse(JSON.parse(data.Payload as string).body));
+        });
+      });
+  }
```
vponline, line 290: The `spawn` function is basically copied from Airnode, but I was wondering if you saw a point to use the `InvocationType: 'Event'` field if we don't need to keep the parent lambda running until all processing finishes.

src/handlers/psp.ts
```diff
@@ -349,8 +330,16 @@ const submitTransactions = async (state: State) => {
...
+  const providerSponsorPromises = providerSponsorSubscriptions.map(async (providerSponsorSubscription) =>
+    spawn({
+      providerSponsorSubscription,
+      baseLogOptions: baseLogOptions,
+      type: config.nodeSettings.cloudProvider.type,
+    })
+  );
+
+  //TODO: do we want to handle lambda promise rejections separately. e.g. Promise.allSettled().catch()
+  await Promise.all(providerSponsorPromises);
```
vponline, line 342: I was thinking `Promise.allSettled` could be used here instead of `Promise.all` here if we want to wait for some return values or errors?

src/handlers/psp.ts
```diff
@@ -249,68 +250,48 @@ const initializeProviders = async (state: State): Promise<State> => {
...
+        // Uses the current region by default
+        const lambda = new AWS.Lambda();
+
+        // TODO: do we want to use the airnode format for FunctionName
+        // AWS doesn't allow uppercase letters in lambda function names
+        // const resolvedName = `airkeeper-${providerSponsorSubscription.sponsorAddress}-run`;
+
+        const options = {
+          FunctionName: 'process-subscriptions',
+          Payload: JSON.stringify({ providerSponsorSubscription, baseLogOptions }),
```
vponline, line 277: The `airnodeWallet.mnemonic.phrase`  is lost here once this is stringified so I'll update the process-suscriptions lambda to load the config file.

src/handlers/psp.ts
amarthadan: I don't think we should do that. I'd actually say that we want to make sure that the lambdas are done within a certain time limit. We still need to be done with the main lambda invocation within 60 seconds so the next run can start. And I think we're doing something like that on Airnode as well. I think the spawned lambdas are wrapped in a timeout.

src/handlers/psp.ts
amarthadan: I think that makes sense. @acenolaza @aquarat Any objections?

serverless.yml
```diff
@@ -24,3 +29,5 @@ functions:
     timeout: 60
     events:
       - schedule: rate(1 minute)
+  process-subscriptions:
+    handler: dist/src/handlers/process-subscriptions.handler
```
amarthadan, line 33: You should definitely add a timeout. Not sure how much though :smile: If the API call processing will take up to 40s this should be done within the remaining 20s. So, let's say 15s to have some time for processing on the main lambda.

.eslintrc.js
amarthadan: Take a look at the AWS handler in `airnode-deployer`. If you prefix the type with the namespace you apparently don't need to import anything.

amarthadan (changes requested):

.eslintrc.js
vponline: Interesting! It turns out we don't need these types at all though.

src/handlers/process-subscriptions.ts
```diff
@@ -0,0 +1,70 @@
...
+    transactionCount
+  );
+
+  logs.forEach(([logs, data]) => {
+    const subscriptionLogOptions = buildLogOptions('additional', { subscriptionId: data.id }, sponsorWalletLogOptions);
+    utils.logger.logPending(logs, subscriptionLogOptions);
+  });
+};
+
+export const handler = async (event: APIGatewayEvent | { body: string }) => {
```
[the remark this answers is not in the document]
vponline: Fixed :+1:

src/handlers/psp.ts
```diff
@@ -249,68 +250,48 @@ const initializeProviders = async (state: State): Promise<State> => {
...
+      return new Promise((resolve, reject) => {
+        // Uses the current region by default
+        const lambda = new AWS.Lambda();
+
+        // TODO: do we want to use the airnode format for FunctionName
+        // AWS doesn't allow uppercase letters in lambda function names
+        // const resolvedName = `airkeeper-${providerSponsorSubscription.sponsorAddress}-run`;
+
+        const options = {
+          FunctionName: 'process-subscriptions',
```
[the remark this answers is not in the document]
vponline: Yep it works (but the name field needs to be in the **yml** file). I changed it to the more standard version though as I think this is only a serverless framework feature.

src/handlers/process-subscriptions.ts
```diff
@@ -0,0 +1,115 @@
...
+      ...providerSponsorSubscription,
+      providerState: { ...providerSponsorSubscription.providerGroup, airnodeWallet, ...evmProviderState },
+    },
+    baseLogOptions
+  );
+
+  return {
+    statusCode: 200,
+    ok: true,
+    message: `Processing subscriptions for sponsorAddress: ${providerSponsorSubscription.sponsorAddress} has finished`,
```
vponline, line 113: I've moved config loading and  `initializeProvider` inside this function.

src/handlers/psp.ts
```diff
@@ -56,6 +56,27 @@ export const handler = async (_event: any = {}): Promise<any> => {
...
+        providerUrl: chainProvider.url,
+        chainConfig: chain,
+      };
+    })
+  );
+
+  const filteredGroupedProviders = groupedProviders.filter((p) => !isNil(p));
+
+  return filteredGroupedProviders;
+};
```
vponline, line 78: We don't need to initialize providers in the main function anymore, so I simply converted the function to group them with the necessary data.

src/handlers/psp.ts
```diff
@@ -213,121 +236,80 @@ const executeApiCalls = async (state: State): Promise<State> => {
...
+
+          // Reject non-ok results
+          if (!parsedData.ok) {
+            reject(parsedData.message);
+            return;
+          }
+
+          resolve(parsedData.message);
+        });
+      });
```
vponline, line 290: I refactored the `spawn` function to return information from each `process-subscription` handler and resolve/reject accordingly.

src/handlers/process-subscriptions.ts
acenolaza: Just being curious, would that mean that we will be loading the same config file for each provider in the config?

src/handlers/psp.ts
acenolaza: The main goal of `initializeProviders` function was to do tasks getting the block number, etc only once. If that goal remains after moving it to `process-subscriptions` then I'm OK with it.

src/handlers/process-subscriptions.ts
```diff
@@ -0,0 +1,70 @@
...
+    subscriptions,
+    sponsorWallet,
+    voidSigner,
+    transactionCount
+  );
+
+  logs.forEach(([logs, data]) => {
+    const subscriptionLogOptions = buildLogOptions('additional', { subscriptionId: data.id }, sponsorWalletLogOptions);
+    utils.logger.logPending(logs, subscriptionLogOptions);
+  });
```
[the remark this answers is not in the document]
vponline: I changed the name, but please let me know if you think it's better to log in the main function

src/handlers/psp.ts
```diff
@@ -213,121 +236,80 @@ const executeApiCalls = async (state: State): Promise<State> => {
...
+  switch (type) {
+    case 'local':
+      return new Promise((resolve, reject) => {
+        processSubscriptionsHandler({ providerSponsorSubscription, baseLogOptions }).then((data) => {
+          if (!data.ok) {
+            reject(data.message);
+          }
+          resolve(data.message);
+        });
+      });
```
acenolaza, line 260: I was really concerned that we would not be able to run Airkeeper locally after this PR so I'm really happy to see that you are supporting local invocations 🙌🏻

src/handlers/psp.ts
```diff
@@ -213,121 +236,80 @@ const executeApiCalls = async (state: State): Promise<State> => {
...
 
-const initializeProviders = async (state: State): Promise<State> => {
-  const { config, baseLogOptions } = state;
-
-  const airnodeWallet = ethers.Wallet.fromMnemonic(config.nodeSettings.airnodeWalletMnemonic);
-
-  const evmChains = config.chains.filter((chain) => chain.type === 'evm');
-  if (isEmpty(evmChains)) {
-    throw new Error('One or more evm compatible chains must be defined in the provided config');
+export const spawn = ({
```
acenolaza, line 239: I guess it would be nice to move this function into it's own file similar to aws.ts in node package

acenolaza (commented):
Few comments while I checkout the branch to test it locally. Great work Vekil 💪🏻

serverless.yml
```diff
@@ -8,6 +8,11 @@ provider:
   memorySize: 512
   logRetentionInDays: 14
   environment: ${file(load-secrets-env.js)}
+  iamRoleStatements:
+    - Effect: 'Allow'
+      Action:
+        - 'lambda:InvokeFunction'
+      Resource: '*'
```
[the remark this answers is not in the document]
vponline: It's possible to do what Michal described, but specifying lambda-level roles disables the serverless defaults (like cloudwatch policies) so we would need to define these manually as well (more details [here](https://www.serverless.com/framework/docs/providers/aws/guide/iam/#custom-iam-roles)).

So we concluded to define this at the provider level, but restrict the target resource to the `process-subscriptions` lambda.

src/handlers/process-subscriptions.ts
vponline: Yes we are loading it for each provider to get the `airnodeWalletMnemonic`. I figured it's probably not a good idea to send it in the `lambda.invoke` **Payload** :smile:

src/handlers/psp.ts
vponline: Good point, how about a `workers` folder like in Airnode?

src/handlers/process-subscriptions.ts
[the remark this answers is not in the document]
acenolaza: I usually call that variable `logsData` or `result` when it's wrapped by a promise-utils `go()` or `goSync()` but that's just my own convention.

src/handlers/psp.ts
acenolaza: No objections and we could even log errors that might occur for each provider/sponsor promise.

One thing I find weird here is that variable name `providerSponsorSubscription` which includes all subscriptions for a provider/sponsor pair. I feel like `providerSponsorSubscriptions` should be changed to `providerSponsorSubscriptionsList` or `providerSponsorSubscriptionsArray`

src/evm/process-sponsor-wallet.ts
```diff
@@ -19,6 +19,7 @@ export const processSponsorWallet = async (
 
   // Keep track of nonce outside of the loop in case there is an invalid subscription and its nonce is skipped
   let nextNonce = transactionCount;
+
```
acenolaza, line 22: I don't know if this comment is relevant in this PR but it seems that we've missed to cover lines 37-38 in the tests after last refactor.

serverless.yml
```diff
@@ -2,12 +2,17 @@ service: airkeeper
 
 provider:
   name: aws
-  # region: ${opt:region}
-  # stage: ${opt:stage}
+  region: ${env:REGION}
+  stage: ${env:STAGE}
```
acenolaza, line 6: Is this now being used?

src/evm/process-sponsor-wallet.ts
vponline: Do you mean testing a skipped invalid subscrptions? If so, I though the changes to [this test](https://github.com/api3dao/airkeeper/pull/50/files#diff-360f756b43e60a608f85b831e840b2ea931db0271eed2c3dc2ba021e45433942R202-R239) would be enough. Although I now realize I didn't update the name of the test to match.

serverless.yml
vponline: I set them in the provider so they could be used in the lambda IAM role:
`'arn:aws:lambda:${self:provider.region}:${aws:accountId}:function:airkeeper-${self:provider.stage}-process-subscriptions'`

> Tested it locally and it seems to be working fine but I must admit that it is a bit annoying to have to set valid AWS creds in order to run Airkeeper locally

I think we could add defaults for dev values. This should work:
```suggestion
  region: ${env:REGION, 'us-east-1'}
  stage: ${env:STAGE, 'dev'}
```

src/handlers/psp.ts
vponline: Agreed @acenolaza  I renamed:
providerSponsorSubscription -> providerSponsorSubscriptions (the type for this was already called ProviderSponsorSubscriptions)
providerSponsorSubscriptions -> providerSponsorSubscriptionsArray

src/workers/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  type: 'local' | 'aws' | 'gcp';
+  stage: string;
+}): Promise<string> => {
+  switch (type) {
+    case 'local':
+      return local.spawn({ providerSponsorSubscriptions, baseLogOptions, stage });
+    case 'aws':
+      return aws.spawn({ providerSponsorSubscriptions, baseLogOptions, stage });
+    case 'gcp':
+      return new Promise((_, reject) => reject('GCP is not supported'));
```
vponline, line 23: I moved the `spawn` functions to a `workers` folder and renamed providerSponsorSubscriptions.

I had to add this case so that the return value from `spawn` didn't include `undefined`.

src/handlers/psp.ts
```diff
@@ -349,8 +267,40 @@ const submitTransactions = async (state: State) => {
...
+      }
+      return acc;
+    },
+    { fulfilled: [], rejected: [] }
+  );
+
+  utils.logger.info(
+    `Processed subscriptions; success: ${groupedResults.fulfilled.length}, failure: ${groupedResults.rejected.length}`,
+    baseLogOptions
+  );
```
amarthadan, line 303: I don't think this is giving that much info tbh. It's a nice statistic but I would much rather know what subscriptions failed and why.

I think you should log the messages you're currently returning from lambda but not doing anything with. But I guess that can be done in the lambda and not here.

src/workers/local.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const spawn = async ({ providerSponsorSubscriptions, baseLogOptions }: WorkerParameters): Promise<string> =>
+  new Promise((resolve, reject) => {
+    processSubscriptionsHandler({ providerSponsorSubscriptions, baseLogOptions }).then((data) => {
+      if (!data.ok) {
+        reject(data.message);
+      }
+      resolve(data.message);
+    });
+  });
```
amarthadan, line 12: Do we even need this file and the local `spawn` version? Why don't we just call the `processSubscriptionsHandler` directly for the `local` case?

src/workers/aws.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const spawn = async ({
+  providerSponsorSubscriptions,
+  baseLogOptions,
+  stage,
+}: WorkerParameters): Promise<string> =>
+  // lambda.invoke is synchronous so we need to wrap this in a promise
+  new Promise((resolve, reject) => {
+    // Uses the current region by default
+    const lambda = new AWS.Lambda({ endpoint: 'http://localhost:3002' });
```
amarthadan, line 12: Why the localhost endpoint?

amarthadan (changes requested):

amarthadan:
> Tested it locally and it seems to be working fine but I must admit that it is a bit annoying to have to set valid AWS creds in order to run Airkeeper locally sweat_smile

I haven't tried to run it but why would we need that?

src/evm/process-sponsor-wallet.ts
acenolaza: I mean when condition check doesn't result in a failure but instead returns false.

src/handlers/psp.ts
```diff
@@ -349,8 +267,40 @@ const submitTransactions = async (state: State) => {
...
   );
 
-  // TODO: start new lambdas for each providerSubscriptions array element
-  await processSubscriptions(providerSponsorSubscriptions, baseLogOptions);
+  const providerSponsorPromises = providerSponsorSubscriptionsArray.map(async (providerSponsorSubscriptions) =>
+    spawn({
+      providerSponsorSubscriptions,
+      baseLogOptions: baseLogOptions,
+      type: config.nodeSettings.cloudProvider.type,
+      stage: config.nodeSettings.stage,
```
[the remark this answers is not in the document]
vponline: Very good point :smile:

src/handlers/process-subscriptions.ts
```diff
@@ -0,0 +1,115 @@
...
+  utils.logger.logPending(logs, providerLogOptions);
+  if (isNil(evmProviderState)) {
+    const message = 'Failed to initialize provider';
+    utils.logger.warn(message, providerLogOptions);
+
+    return {
+      statusCode: 500,
+      ok: false,
+      message: `Failed to initialize provider: ${providerSponsorSubscriptions.providerGroup.providerName} for chain: ${providerSponsorSubscriptions.providerGroup.chainId}`,
+    };
```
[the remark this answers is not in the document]
vponline: I just figured I'd use the same format as the returned value from the main PSP lambda, but ended up simplifying the return data to just `message` as I didn't see a point to deal with stringifying and parsing unless we want to return more detailed data.
```ts
const response = {
    ok: true,
    data: { message: 'PSP beacon update execution has finished' },
  };
  return { statusCode: 200, body: JSON.stringify(response) };
```

The `message` is actually just returned from the `spawn` function, but you're right about `statusCode` not being used. I wasn't quite sure what data we want or need to return/log. Any preference to what should be returned instead?

src/workers/aws.ts
vponline: I forgot to remove this (I was testing with serverless-offline).

src/workers/local.ts
vponline: You're right, we can do it that way as well :+1:

src/handlers/process-subscriptions.ts
```diff
@@ -0,0 +1,115 @@
...
+    utils.logger.error(airnodeConfig.error.message);
+    throw airnodeConfig.error;
+  }
+
+  const airnodeWallet = ethers.Wallet.fromMnemonic(airnodeConfig.data.nodeSettings.airnodeWalletMnemonic);
+
+  // Initialize provider specific data
+  const [logs, evmProviderState] = await initializeProvider(
+    providerSponsorSubscriptions.providerGroup.chainConfig,
+    providerSponsorSubscriptions.providerGroup.providerUrl || ''
```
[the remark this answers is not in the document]
vponline: Good point, I'll update it like this

acenolaza:
> > Tested it locally and it seems to be working fine but I must admit that it is a bit annoying to have to set valid AWS creds in order to run Airkeeper locally sweat_smile
> 
> I haven't tried to run it but why would we need that?

My guess is because of trying to get the `${aws:accountId}` in the serverless.yml config file.

amarthadan:
> > > Tested it locally and it seems to be working fine but I must admit that it is a bit annoying to have to set valid AWS creds in order to run Airkeeper locally sweat_smile
> > 
> > 
> > I haven't tried to run it but why would we need that?
> 
> My guess is because of trying to get the `${aws:accountId}` in the [serverless.yml config file](https://github.com/api3dao/airkeeper/blob/3d237643473f40e411e212439b7154b9b903e86b/serverless.yml#L15).

:thinking: I would expect serverless to ignore role settings when running locally but you might be right. @vponline can you test that?

src/handlers/psp.ts
amarthadan: > I think all subscription specific logs are handled in process-subscriptions.ts [here](https://github.com/api3dao/airkeeper/pull/53/files#diff-62a0c95e58afb4957b9bf3376a5ccf9c38745a1946518a36d00fb0bfd0fe06b2R62-R65), right?

Yes

> How about doing it like this to log possibly rejected promises (i.e. errors from the lambda)?

But wouldn't be better to log it in the lambdas instead of returning it and logging it here?

vponline:
> I would expect serverless to ignore role settings when running locally but you might be right. @vponline can you test that?

@amarthadan  It seems that using the serverless variable `aws:accountId` it does require it to be set locally, but using the [AWS pseudo-parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html) `AWS::AccountId` it works without.

@acenolaza If you have a chance, could you also please try this and let me know that it works for you as well.

src/handlers/psp.ts
vponline: > But wouldn't be better to log it in the lambdas instead of returning it and logging it here?

Yes for the successful cases, but I think we would still need to log (unhandled) invoke lambda errors from the spawn function in them main lambda: 

```ts
lambda.invoke(options, (err, data) => {
      // Reject invoke and (unhandled) handler errors
      if (err || data.FunctionError) {
        reject(err || data.FunctionError);
        return;
      }
...
```

Or did you have something else in mind?

amarthadan: Yes, but this is different from what you wrote above. I think we should log success and expected failures in the lambda but of course, in case of the lambda invocation failure, we have to log that here.

vponline: > Yes, but this is different from what you wrote above. I think we should log success and expected failures in the lambda but of course, in case of the lambda invocation failure, we have to log that here.

Just to make sure one more time :smile:, these are the changes:
1. we remove the `result.status === 'fulfilled'` block from the main lambda and leave the `rejected` block.
2. we log `Processing subscriptions for sponsorAddress.... finished` in the `process-subscriptions` lambda (and remove the `message` and `statusCode` fields as those are not needed anymore).

amarthadan: Yes, let's do it that way :slightly_smiling_face:

acenolaza:
> > I would expect serverless to ignore role settings when running locally but you might be right. @vponline can you test that?
> 
> @amarthadan It seems that using the serverless variable `aws:accountId` it does require it to be set locally, but using the [AWS pseudo-parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html) `AWS::AccountId` it works without.
> 
> @acenolaza If you have a chance, could you also please try this and let me know that it works for you as well.

Works for me 💪🏻

package.json
```diff
@@ -29,7 +29,7 @@
     "sls:invoke:rrp-beacon-update": "sls invoke --function rrp-beacon-update",
     "sls:invoke-local:rrp-beacon-update": "yarn build && sls invoke local --function rrp-beacon-update",
     "sls:invoke:psp-beacon-update": "sls invoke --function psp-beacon-update",
-    "sls:invoke-local:psp-beacon-update": "yarn build && sls invoke local --function psp-beacon-update",
+    "sls:invoke-local:psp-beacon-update": "yarn build && sls invoke local --function psp-beacon-update --param='provider=local'",
```
amarthadan, line 32: Oh, I guess this is the `param:provider` I was asking about :smile: Yeah, that makes sense :slightly_smiling_face:

amarthadan (changes requested):

src/workers/aws.ts
```diff
@@ -0,0 +1,34 @@
...
+
+      const parsedData: AWSHandlerResponse = JSON.parse(data.Payload as string);
+
+      // Reject non-ok results
+      if (!parsedData.ok) {
+        reject(parsedData);
+        return;
+      }
+
+      resolve(parsedData);
```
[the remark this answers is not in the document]
vponline: Yes that does make sense :+1:

src/handlers/psp.ts
```diff
@@ -391,8 +328,22 @@ const submitTransactions = async (state: State) => {
...
   );
 
-  // TODO: start new lambdas for each providerSubscriptions array element
-  await processSubscriptions(providerSponsorSubscriptions, baseLogOptions);
+  const providerSponsorPromises = providerSponsorSubscriptionsArray.map(async (providerSponsorSubscriptions) =>
+    spawn({
+      providerSponsorSubscriptions,
+      baseLogOptions: baseLogOptions,
+      type: process.env.CLOUD_PROVIDER as 'local' | 'aws' | 'gcp',
+      stage: process.env.STAGE!,
```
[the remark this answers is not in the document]
vponline: You are right! I'll fix this by setting in under psp-beacon-update lambda environment variables

src/handlers/process-subscriptions.ts
```diff
@@ -0,0 +1,126 @@
...
+  const contracts = Object.entries(providerSponsorSubscriptions.providerState.chainConfig.contracts).reduce(
+    (acc, [contractName, contractAddress]) => {
+      if (isNil(abis[contractName])) {
+        return acc;
+      }
+      return { ...acc, [contractName]: new ethers.Contract(contractAddress, abis[contractName], provider) };
+    },
+    {}
+  );
+  const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
```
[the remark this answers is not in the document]
vponline: I moved this out and called the function `reinitializeProvider`, but I guess it would be more accurate to refactor the current `initializeProvider` function to be called `initializeEvmState` (or something similar) and call this function `initializeProvider`. Do you think this is necessary?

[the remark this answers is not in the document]
amarthadan: Sure, I agree with that :+1:

amarthadan (approved):
I'd say we're there :muscle: :tada: :smile:

PR #54, opened 2022-03-29
Bec254 improve api calls

acenolaza:
These changes will allow Airkeeper to:
1. retry each individual API call indefinitely on failure with a random backoff or until a 10 sec timeout is reached
2. all API calls will be triggered in parallel with a shared timeout of 40 sec
3. each API call will store its result in a shared variable in memory and Airkeeper will try to continue processing with as many results were retrieved from API.

Note 1: when timeouts are reached, background running processes are not killed there I added those boolean flags to avoid having any unhandled exception when there's a timeout.

Note: I ran into issue when trying to accomplish the same using promise-utils package `go()` function and since this function needs to be changed to support more AttemptOptions params then I thought it would be good to work on those changes then publish a new promise-utils package and finally switch to using promise-utils in here.

### DIFF
--- a/src/handlers/psp.ts
+++ b/src/handlers/psp.ts
@@ -171,15 +171,16 @@ const executeApiCalls = async (state: State): Promise<State> => {
   const wait = (ms: number) => new Promise((res) => setTimeout(res, ms));
   const doRequestWithRetries = async (
     fn: () => Promise<CallApiResult>,
-    minDelay = 1_000,
-    maxDelay = 5_000
+    minDelay = 200,
+    maxDelay = 2000
   ): Promise<CallApiResult> => {
-    try {
-      return await fn();
-    } catch (e) {
-      await wait(Math.floor(Math.random() * (maxDelay - minDelay + 1) + minDelay));
+    const goResult = await promise.go(fn());
+    if (!goResult.success) {
+      const delay = Math.floor(Math.random() * (maxDelay - minDelay + 1) + minDelay);
+      await wait(delay);
       return await doRequestWithRetries(fn, minDelay, maxDelay);
     }
+    return goResult.data;
   };
 
   let hasApiCallWrapperTimedOut = false;
@@ -188,11 +189,12 @@ const executeApiCalls = async (state: State): Promise<State> => {
   const apiValuePromises = groupedSubscriptions.map(async ({ subscriptions, template, endpoint }) => {
     const apiCallParameters = abi.decode(template.templateParameters);
 
-    let hasApiCallTimedOut = false;
     const result = await doRequestWithRetries(async () => {
       if (hasApiCallWrapperTimedOut) {
         return [[], { templateId: template.id, apiValue: null, subscriptions }] as CallApiResult;
       }
+
+      let hasApiCallTimedOut = false;
       const result = await promise.go(
         async () => {
           try {
--- a/src/handlers/psp.ts
+++ b/src/handlers/psp.ts
@@ -1,8 +1,7 @@
 import * as abi from '@api3/airnode-abi';
-import * as node from '@api3/airnode-node';
 import * as utils from '@api3/airnode-utilities';
 import { goSync } from '@api3/promise-utils';
-import { retry } from '@lifeomic/attempt';
+import { AttemptOptions, retry } from '@lifeomic/attempt';
 import { ethers } from 'ethers';
 import groupBy from 'lodash/groupBy';
 import isEmpty from 'lodash/isEmpty';
@@ -12,14 +11,15 @@ import { loadAirkeeperConfig, loadAirnodeConfig, mergeConfigs } from '../config'
 import { getSponsorWalletAndTransactionCount, initializeProvider, processSponsorWallet } from '../evm';
 import { buildLogOptions } from '../logger';
 import {
+  CallApiResult,
+  CheckedSubscription,
   Config,
   EVMProviderState,
   GroupedSubscriptions,
   Id,
+  ProviderSponsorSubscriptions,
   ProviderState,
   State,
-  CheckedSubscription,
-  ProviderSponsorSubscriptions,
 } from '../types';
 import { Subscription } from '../validator';
 import { shortenAddress } from '../wallet';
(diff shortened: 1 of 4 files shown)


src/handlers/psp.ts
```diff
@@ -176,29 +177,67 @@ const initializeState = (config: Config): State => {
...
-          templateId: string;
-          apiValue: ethers.BigNumber | null;
-          subscriptions: Id<Subscription>[];
-        }>;
-      },
-      { timeoutMs: TIMEOUT_MS, retries: RETRIES }
-    );
-  });
-  const responses = await Promise.all(apiValuePromises);
+    let hasApiCallTimedout = false;
```
Siegrift, line 199:
```suggestion
    let hasApiCallTimedOut = false;
```

Siegrift (changes requested):
The logic seems a bit too complex and I guess I understand the problem. I think it can be simplified like this:

1. For every request do
2. Create a `go` with timeout 40 seconds to make sure we have some response
3. Inside the callback, use `go` to make the api request with 10s timeout with a random sleep in a while cycle until you have a valid response 
4. Use `Promise.all(promises)` (with no timeout) and filter successfull responses (and log the failed ones)

amarthadan (commented):
Ok, I think we're actually on the right track here. Once the code is changed to use the `go` function and we get rid of the try-catch blocks it will be much nicer.

I don't really agree with @Siegrift's suggestion to wrap each API call into a 40s promise. I think it's kinda confusing and the result is pretty much the same.

This is what I would suggest:

Let's get rid of the try-catch blocks and use the `go` function right away. I know we're missing some options but based on the `attempt` library documentation they won't really help us with the random back-off anyway. There's a `jitter` option but the math behind it seems weird to me. Correct me if I'm wrong. So, I would suggest retrying and doing the random backoff manually (we might implement that to the `promise-utils` library though. We can use `setTimeout` with a random value from a range for that.

We can also get rid of the `hasApiCallTimedout` flag. We've spoken with @acenolaza on Slack and there was some misunderstanding about what the 10s timeout is for but it should be fine now :slightly_smiling_face:

We will use the 40s retry for the `Promise.all` in the end, setting the flag to `true` once timeouted. But I'd say we need to check for that flag not only in case of the API call fails but also in case it succeeds because we want to avoid manipulating the `responses` array once we start iterating it. So, if some API call finishes after the 40s timeout (either sucessufly or not) we just log the result and not add it to `responses` array.

What do you guys say, does this make sense?

Siegrift:
> What do you guys say, does this make sense?

Not to me, but I might be missing something. I think that if the 40s promise fails and you set the cleaning flag to true, you will then need to wait again until all of the promises either finish or time out. That works, but I find that ugly and we will actually have to wait for 50 seconds in worst case (or set the time limit to 30s instead of 40s)

Here's how I'd implement it - hopefully it works:
https://github.com/api3dao/airkeeper/commit/ff26f2f872e00a38323ecdff8326cd675928b4f2

src/handlers/psp.ts
```diff
@@ -166,32 +165,65 @@ const initializeState = (config: Config): State => {
...
+        async () => {
+          try {
+            return await callApi(config, endpoint, apiCallParameters);
+          } catch (err) {
+            if (hasApiCallTimedout) {
+              return [
+                [utils.logger.pend('ERROR', '' + err)],
+                { templateId: template.id, apiValue: null, subscriptions },
+              ] as CallApiResult;
+            }
```
acenolaza, line 205: I haven't been able to get rid of this flag because of an edge case where the callApi() call has timed out (more that 10sec) and the callApi() throws an error after this time out. So I use this flag to handle any error that might occur after timing out.

acenolaza (commented):
What do you guys think of this alternative? It's uses recursion for the retry instead of a while loop and go function calls to trigger both timeouts (individual API call of 10secs and 40secs for the whole list of API call promises).

src/handlers/psp.ts
Siegrift: @amarthadan showed me an example that, but the issue is actually inside the `go` function. 

Specifically, it's because of a bug inside the `attemp` library here https://github.com/lifeomic/attempt/blob/master/src/index.ts#L187 where it calls on error which throws here https://github.com/lifeomic/attempt/blob/master/src/index.ts#L146

This is problematic, because it breaks the promise chain and when you do `Promise.resolve(somePromise)` it returns `somePromise` which is not handled there. The fix would be to add a `catch` clause after it. But honestly, promises should not be used like this, you should strive to make linear promise chains which are easy to reason about, which is not the case of that library. I'll take care of `promise-utils` and let Burak publish it. 

As of now, I think having something that works is fine. I'd still prefer [this implementation](https://github.com/api3dao/airkeeper/commit/ff26f2f872e00a38323ecdff8326cd675928b4f2), but I'd just create a JIRA issue and we can update it after there is a new version of go utils. WDYT?

Siegrift (approved):
I think the current handling should work and a proper fix requires more changes (and in different package) so I'd suggest keeping this implementation and revisiting it later.

src/handlers/psp.ts
```diff
@@ -166,32 +165,65 @@ const initializeState = (config: Config): State => {
...
+        return [[], { templateId: template.id, apiValue: null, subscriptions }] as CallApiResult;
+      }
+      const result = await go(
+        async () => {
+          try {
+            return await callApi(config, endpoint, apiCallParameters);
+          } catch (err) {
+            if (hasApiCallTimedout) {
+              return [
+                [utils.logger.pend('ERROR', '' + err)],
```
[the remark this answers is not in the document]
acenolaza: I can add better handling for that case 👍🏻

acenolaza:
> I think the current handling should work and a proper fix requires more changes (and in different package) so I'd suggest keeping this implementation and revisiting it later.

Thanks @Siegrift for looking into this. I feel like I have spent way too much time on this and even though it is not perfect...it does the job 😅 We can keep improving it after the new promise-utils package is updated like you mentioned.

So if @amarthadan is OK with all this then I'll just go ahead and merge it after adding those minor fixes from your comments and maybe some other minor cleanup.

src/handlers/psp.ts
```diff
@@ -169,29 +168,70 @@ const initializeState = (config: Config): State => {
 const executeApiCalls = async (state: State): Promise<State> => {
   const { config, baseLogOptions, groupedSubscriptions } = state;
 
-  const apiValuePromises = groupedSubscriptions.map(({ subscriptions, template, endpoint }) => {
+  const wait = (ms: number) => new Promise((res) => setTimeout(res, ms));
+  const doRequestWithRetries = async (
+    fn: () => Promise<CallApiResult>,
+    minDelay = 1_000,
+    maxDelay = 5_000
```
amarthadan, line 175: I'd say let's try a range from 200ms to 2s.

src/handlers/psp.ts
```diff
@@ -169,29 +168,70 @@ const initializeState = (config: Config): State => {
...
+  ): Promise<CallApiResult> => {
+    try {
+      return await fn();
+    } catch (e) {
+      await wait(Math.floor(Math.random() * (maxDelay - minDelay + 1) + minDelay));
+      return await doRequestWithRetries(fn, minDelay, maxDelay);
+    }
```
amarthadan, line 182: Can't we use the `go` function here as well? Instead of the try-catch block I mean.

src/handlers/psp.ts
```diff
@@ -169,29 +168,70 @@ const initializeState = (config: Config): State => {
...
+        async () => {
+          try {
+            return await callApi(config, endpoint, apiCallParameters);
+          } catch (err) {
+            if (hasApiCallTimedOut) {
+              return [
+                [utils.logger.pend('ERROR', err instanceof Error ? err.message : String(err))],
+                { templateId: template.id, apiValue: null, subscriptions },
+              ] as CallApiResult;
+            }
```
amarthadan, line 206: I still think this doesn't work in all cases. Imagine that the first attempt will timeout, meaning that the `callApi` is still running. We set the `hasApiCallTimedOut` to true and start the second attempt. If that fails (within the timeout) it will be falsely considered as a failure of the first attempt after the timeout. Feel free to correct me if I'm wrong.

amarthadan (approved):
I'm ok with merging this for now but the issue I described needs to be fixed before the production release. We can discuss what to do about it on Slack.

src/handlers/psp.ts
acenolaza: We probably can. I'll make the change and test it out.

src/handlers/psp.ts
acenolaza: The scenario makes sense. Let me try to reproduce it locally and see what solution can be applied if true.

amarthadan: I don't think it's actually fixable in the current state, that's why I want to address it once we have the library fixed.

acenolaza: I see. The thing is that I'm almost 💯 sure that I did test that scenario and it was working but the logic of what you said makes sense so I want to double check if it's an issue or not.

acenolaza: Ok you were right @amarthadan. I just tested it and after timeout, if the next retry fails then we return null and stop retrying even though we want to keep retrying until we hit the 40 sec limit for all API requests.

I might be missing something but I think the problem is that I'm initializing the `hasApiCallTimedOut` outside the retry callback function. If we move it inside in between lines 195-196 then we would throw after the first timeout and on the next retry the flag would be false and keep retrying. I did a test of that change and it seems to be working but I might be missing other edge cases...

acenolaza: Done! Ran a few tests with random timeouts + random failures and it seems to behave as expected

PR #55, opened 2022-04-06
Add E2E tests for PSP

vponline:
This adds E2E tests for PSP beacon updates.


package.json
```diff
@@ -19,6 +19,13 @@
     "clean": "rimraf -rf ./dist *.tgz",
     "compile": "tsc -p tsconfig.json",
     "compile:build": "tsc -p tsconfig-build.json",
+    "dev:clean": "rimraf -rf ./logs",
+    "dev:list": "pm2 list",
+    "dev:stop": "pm2 stop ecosystem.config.js",
+    "dev:delete": "pm2 delete all",
+    "dev:eth-node": "hardhat node --config src/test/hardhat.config.ts",
```
vponline, line 26: I placed the hardhat config inside **test/** because it was causing a problem with **airnode-protocol** contract compilation in the postinstall script, but this can be moved to the root folder once we get the new Airnode packages  released.

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,185 @@
...
+        templateId,
+        encodedBeaconUpdateSubscriptionConditions,
+        airnodeWallet.address,
+        roles.sponsor.address
+      );
+  });
+
+  it('updates the beacon successfully', async () => {
+    jest.spyOn(config, 'loadAirnodeConfig').mockImplementation(() => airnodeConfig as any);
+    jest.spyOn(config, 'loadAirkeeperConfig').mockImplementation(() => airkeeperConfig as any);
```
vponline, line 128: I wasn't able to make these config mocks work with `mockReadFileSync` as the **airnodeConfig** loading keeps failing in process-subscriptions.

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,189 @@
...
+import { mockReadFileSync } from '../mock-utils';
+import { ethers } from 'ethers';
+import * as abi from '@api3/airnode-abi';
+import * as node from '@api3/airnode-node';
+import * as psp from '../../handlers/psp';
+import * as config from '../../config';
+import { buildAirnodeConfig, buildAirkeeperConfig, buildLocalConfig } from '../config/config';
+import { PROTOCOL_ID_PSP } from '../../constants';
+
+jest.setTimeout(30_000);
```
vponline, line 12: There's a [bug in Jest >v.27](https://github.com/facebook/jest/issues/11607) where a **setTimeout** doesn't work correctly inside of a describe block :smile:

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  it('updates the beacon successfully', async () => {
+    jest
+      .spyOn(config, 'loadAirnodeConfig')
+      .mockImplementationOnce(() => airnodeConfig as any)
+      .mockImplementationOnce(() => airnodeConfig as any);
+    jest.spyOn(config, 'loadAirkeeperConfig').mockImplementationOnce(() => airkeeperConfig as any);
+    const res = await psp.handler();
+
+    expect(dapiServer).toBeDefined();
```
amarthadan, line 135: Why is this needed? This is initialized outside the test itself, isn't it?

amarthadan (commented):
I quite like this :+1:

Could we maybe add some test(s) for unsuccessful updates? What are the cases when the update can fail?

Also, since we have that retry mechanism for API calls, can we test that too?

amarthadan:
> Do we need tests for multiple subscriptions? I might need help with setting those up.

I guess that would be nice. What's the problem with that?

amarthadan:
@acenolaza @Siegrift Can you take a look at this as well?

amarthadan:
Oh, one thing I forgot, can you update README with instructions on how to run the e2e tests?

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,189 @@
...
+    expect(res).toEqual({
+      statusCode: 200,
+      body: JSON.stringify({ ok: true, data: { message: 'PSP beacon update execution has finished' } }),
+    });
```
acenolaza, line 139: I'm not sure that this is enough to determine if the beacon was successfully updated. I think we should be spying on the DapiServer functions in order to check that the expected ones where actually called with the expected args. WDYT?

acenolaza (changes requested):
I was trying my best to avoid having to add hardhat as a dep but I don't see any other way around when it comes to e2e. Unless we create a different repo that deploys an Airkeeper using the docker image etc. but that might be too much work at the moment. 

The good thing is that now we could reference the contracts from lets say operations repo and build them locally to get the ABI instead of having them hardcoded in different [files](https://github.com/api3dao/airkeeper/tree/main/scripts/artifacts).

src/test/e2e/psp.feature.ts
Siegrift: It would be nice to add that as a comment (with a link to that issue).

src/test/e2e/psp.feature.ts
Siegrift: Can we instead of spying just read the value and see if it's updated? (just asking, not sure if it would work)

vponline: Yes agreed, great suggestion, I'll add :+1:

src/test/mock-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// Declare originalFs outside of mockReadFileSync to prevent infinite recursion errors in mockReadFileSync.
+const originalFs = fs.readFileSync;
+
+/**
+ * Mocks the fs library if the file path includes the specified file path substring
+ * and otherwise returns the original content.
+ */
+export const mockReadFileSync = (filePathSubstr: string, mockValue: string) => {
+  return jest.spyOn(fs, 'readFileSync').mockImplementationOnce((...args) => {
```
Siegrift, line 13: This (unforunately) needs to be mockImplementation because the mock implementation will get consumed if you try to log something before the file is loaded.

I've checked codesearch (https://cs.github.com/api3dao/airseeker/blob/ee2bf2db512c0e69915281af6daca2406d910b9f/src/config.test.ts?q=%22const+mockReadFileSync%22#L12) and we have the same problem in airseeker.

vponline: Oh right, I think we had something similar in Airnode :sweat_smile:

src/test/e2e/psp.feature.ts
vponline: It is not needed :smile:  (I was having problems with hardhat version 2.9.2 so was confirming that the contract was actually deployed). I'll remove

package.json
vponline: hmm it seems to be required, otherwise it doesn't run the network

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,263 @@
...
+  let airnodeProtocolFactory: ContractFactory;
+  let airnodeProtocol: Contract;
+  let dapiServerAbi;
+  let dapiServerFactory: ContractFactory;
+  let dapiServer: Contract;
+
+  beforeEach(async () => {
+    //Reset the local hardhat network state for each test to keep the deployed Airnode and DapiServer contract addresses
+    //the same as the config files
+    await hre.network.provider.send('hardhat_reset');
```
vponline, line 50: This took me way too long to figure this out, but there were a couple issues with the original setup:

1. The (dapiServer, airnodeProtocol) were re-deployed for each test and since the hardhat network was running, those had different addresses for each test. Then the **DapiServer** and **AirnodeRrp** addressed in the config files didn't match and the values were not updated (even though the handler finishes correctly).

I had to make a few modifications to get this hardhat_reset to work (namely copy the hardhat.config.ts to the root after install), otherwise I got an error that he project 'was not a hardhat project'.

2. `mockReadFileSync` does not seem to be working for this (even when I change it to mockImplementation), but it errors out saying that the airnode config cannot be found. So I changed the loadConfig functions to be mocked instead.

3. I tried to derive the **beaconid** with the airnodeProtocol address and the templateId, but it ends up giving me a different id than getting it from the dapiServer contract with `subscriptionIdToBeaconId()`. I'm not sure why that is.

package.json
```diff
@@ -14,11 +14,18 @@
     "*.js"
   ],
   "scripts": {
-    "postinstall": "git clone https://github.com/api3dao/airnode.git airnode; cd airnode && git checkout 994d895f37ca871f57224ac1603ea58e79ef071e && yarn run bootstrap && yarn build && cd ..; cp -a airnode/node_modules/@api3/airnode-node/. node_modules/@api3/airnode-node; rm -rf airnode",
+    "postinstall": "git clone https://github.com/api3dao/airnode.git airnode; cd airnode && git checkout 994d895f37ca871f57224ac1603ea58e79ef071e && yarn run bootstrap && yarn build && cd ..; cp -a airnode/node_modules/@api3/airnode-node/. node_modules/@api3/airnode-node; cp -a src/test/hardhat.config.ts ./; rm -rf airnode",
```
vponline, line 17: We need the hardhat file in the root folder so I'm copying there after installing, this can be also removed once we don't need the postinstall script.

vponline:
I've added tests for retries, invalid subscriptions, a second valid subscription, updated the README, checking updated values, and fixed a problem with hardhat. I think all PR improvements are now covered so please feel free to review again :smile:

amarthadan (approved):
See the last comments but otherwise LGTM :+1:

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,353 @@
...
+
+    expect(dapiServerResponseETH[0].toNumber()).toEqual(723.39202 * 1000000);
+    expect(dapiServerResponseBTC[0].toNumber()).toEqual(41091.12345 * 1000000);
+    expect(res).toEqual({
+      statusCode: 200,
+      body: JSON.stringify({ ok: true, data: { message: 'PSP beacon update execution has finished' } }),
+    });
+  });
+
+  it('updates the beacon successfully with one invalid subscription present', async () => {
```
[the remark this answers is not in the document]
vponline: That's a great idea, ill update

src/test/config/config.ts
```diff
@@ -0,0 +1,301 @@
...
+    { type: 'string32', name: 'to', value: 'USD' },
+    { type: 'string32', name: '_type', value: 'int256' },
+    { type: 'string32', name: '_path', value: 'result' },
+    { type: 'string32', name: '_times', value: '1000000' },
+    { type: 'string32', name: 'from', value: 'ETH' },
+  ],
+  threshold: 10,
+});
+
+export const buildLocalConfig2 = () => ({
```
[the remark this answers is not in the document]
vponline: hmm I thought all the subscriptionIds, templateIds etc. were required to be valid? Please correct me if this is not the case for the future

ecosystem.config.js
```diff
@@ -0,0 +1,42 @@
...
+      // Logs
+      log_date_format: 'YYYY-MM-DD HH:mm:ss',
+      error_file: 'logs/web-api.log',
+      out_file: 'logs/web-api.log',
+      merge_logs: true,
+
+      // Watching
+      watch: ['src'],
+      watch_delay: 500,
+      ignore_watch: ['node_modules'],
```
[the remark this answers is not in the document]
vponline: Yes, it's not really needed

src/test/config/config.ts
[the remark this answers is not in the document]
amarthadan: They do afaik. Why do you ask? I just don't understand what it has to do with what I wrote.

src/test/e2e/psp.feature.ts
Siegrift: I am not sure about the 3), that seems like a thing which might bite us in production. Can you maybe check why is that? (not necessarily in this PR)

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,353 @@
...
+    const airnodeWallet = hre.ethers.Wallet.fromMnemonic(localConfig.airnodeMnemonic);
+    const airnodePspSponsorWallet = node.evm
+      .deriveSponsorWalletFromMnemonic(localConfig.airnodeMnemonic, roles.sponsor.address, PROTOCOL_ID_PSP)
+      .connect(provider);
+    await roles.deployer.sendTransaction({
+      to: airnodePspSponsorWallet.address,
+      value: hre.ethers.utils.parseEther('1'),
+    });
+
+    // Setup ETH Subscription
```
Siegrift, line 103: Mhh, does this comment refer to something?

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,353 @@
...
+      .spyOn(config, 'loadAirnodeConfig')
+      .mockImplementationOnce(() => airnodeConfig as any)
+      .mockImplementationOnce(() => airnodeConfig as any);
+    jest.spyOn(config, 'loadAirkeeperConfig').mockImplementationOnce(() => airkeeperConfig as any);
+    const res = await psp.handler();
+
+    const voidSigner = new hre.ethers.VoidSigner(hre.ethers.constants.AddressZero, provider);
+
+    // ETH price beacon
+    const beaconIdETH = await dapiServer.subscriptionIdToBeaconId(subscriptionIdETH);
```
Siegrift, line 194: Maybe create a helper function for reading a beacon value? That is repeated multiple times in the tests.

Siegrift (approved):
👍 LGTM

src/test/e2e/psp.feature.ts
vponline: Yes so I added this for it to be clear that the following lines are for setting up the ETH subscription (i.e. `buildLocalConfig`)

Maybe these should be renamed to `buildLocalConfigETH` and `buildLocalConfigBTC` to be clear

Siegrift (commented):
One thing, why is the `test` folder inside the `src`? It should be alongside `src`.

src/test/e2e/psp.feature.ts
```diff
@@ -0,0 +1,353 @@
...
+    expect(res).toEqual({
+      statusCode: 200,
+      body: JSON.stringify({ ok: true, data: { message: 'PSP beacon update execution has finished' } }),
+    });
+  });
+
+  it('updates the beacon successfully with one invalid provider present', async () => {
+    jest
+      .spyOn(config, 'loadAirnodeConfig')
+      .mockImplementation(
```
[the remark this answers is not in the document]
vponline: To answer your question, it does work, so `mockImplementation` is setting the default value to be returned after any `mockImplementationOnce` cases are handled

But in this case it's not needed since we want the airnode config to always include the invalid provider, I'll change this to just have `mockImplementation` :+1:

src/test/e2e/psp.feature.ts
acenolaza: @vponline were you able to figure 3 out?

vponline: @Siegrift  @acenolaza  This was my bad, I was trying to derive it with the airnode protocol address instead of the airnode address from the config... :facepalm: I updated `beaconId` to be derived now.

amarthadan:
Is this still waiting for something?

vponline:
> Is this still waiting for something?

Only @acenolaza's final approval :smile:

acenolaza (approved):
Apologies @vponline. It's now approved 👍🏻

PR #56, opened 2022-04-06
Add a test where check conditions returns false

vponline:
This adds a test case to make sure that subscriptions processing continues when one subscription's condition check returns false. (reported [here](https://github.com/api3dao/airkeeper/pull/53#discussion_r839926158)). 

Also updates the test descriptions.


acenolaza (approved):
🏆

PR #57, opened 2022-04-07
Use new promise utils to make api calls

Siegrift:
https://api3dao.atlassian.net/browse/BEC-282

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "sls:invoke-local:psp-beacon-update": "yarn build && sls invoke local --function psp-beacon-update --param='provider=local'",
     "setup:psp-local": "tsc && node scripts/psp-beacon-local-setup.js",
     "test": "jest",
-    "test:debug": "jest \"process-sponsor-wallet.test.ts\"",
     "prepare": "husky install"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "sls:invoke-local:psp-beacon-update": "yarn build && sls invoke local --function psp-beacon-update --param='provider=local'",
     "setup:psp-local": "tsc && node scripts/psp-beacon-local-setup.js",
     "test": "jest",
+    "test:debug": "jest \"process-sponsor-wallet.test.ts\"",
     "prepare": "husky install"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@api3/airnode-node": "^0.5.0",
     "@api3/airnode-protocol": "^0.5.0",
     "@api3/airnode-utilities": "^0.5.0",
-    "@api3/promise-utils": "^0.1.0",
+    "@api3/promise-utils": "^0.2.0",
     "aws-sdk": "^2.1101.0",
     "dotenv-safe": "^8.2.0",
     "ethers": "^5.6.0",
--- a/src/evm/process-sponsor-wallet.test.ts
+++ b/src/evm/process-sponsor-wallet.test.ts
@@ -1,6 +1,5 @@
 import * as node from '@api3/airnode-node';
 import { ethers } from 'ethers';
-import { success } from '@api3/promise-utils';
 import { processSponsorWallet } from './process-sponsor-wallet';
 import { GAS_LIMIT, PROTOCOL_ID_PSP } from '../constants';
 
--- a/src/evm/process-sponsor-wallet.test.ts
+++ b/src/evm/process-sponsor-wallet.test.ts
@@ -1,5 +1,6 @@
 import * as node from '@api3/airnode-node';
 import { ethers } from 'ethers';
+import { success } from '@api3/promise-utils';
 import { processSponsorWallet } from './process-sponsor-wallet';
 import { GAS_LIMIT, PROTOCOL_ID_PSP } from '../constants';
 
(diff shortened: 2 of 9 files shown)


src/evm/process-sponsor-wallet.test.ts
```diff
@@ -245,7 +261,10 @@ describe('processSponsorWallet', () => {
     const fulfillPspBeaconUpdateOnceSpy = jest
       .fn()
       .mockImplementationOnce(fulfillPspBeaconUpdateMock)
-      .mockImplementationOnce((_nameOrSignatureOrSighash: string) => {
+      .mockImplementationOnce(() => {
```
Siegrift, line 264: After the upgrade to new promise utils, this test got broken and I've figured that the function  wasn't retried properly before.

src/handlers/psp.ts
```diff
@@ -200,90 +200,58 @@ const initializeEvmStates = async (state: State): Promise<State> => {
...
 
-    utils.logger.logPending(logs, templateLogOptions);
+  utils.logger.logPending(
+    successfullCalls.flatMap((call) => call[0]),
+    baseLogOptions
+  );
+  utils.logger.logPending(
+    failedCalls.flatMap((call) => call[0]),
+    baseLogOptions
+  );
```
amarthadan, line 240: I understand why we filter successful calls, we use those in `reduce`. But why do we need the failed ones? You're logging them the same way as the successful ones anyway. You can filter the successful ones and log all of them.

src/handlers/psp.ts
```diff
@@ -200,90 +200,58 @@ const initializeEvmStates = async (state: State): Promise<State> => {
...
+      return [
+        [utils.logger.pend('DEBUG', `Retrying API call for templateId ${template.id}: ${goResult.error.message}`)],
+        { templateId: template.id, apiValue: null, subscriptions },
+      ] as CallApiResult;
+    }
+  });
 
-    const templateLogOptions = buildLogOptions('additional', { templateId: data.templateId }, baseLogOptions);
+  const callApiResults = await Promise.all(apiValuePromises);
+  const successfullCalls = callApiResults.filter((call) => call[1].apiValue !== null);
```
amarthadan, line 230:
```suggestion
  const successfulCalls = callApiResults.filter((call) => call[1].apiValue !== null);
```

src/handlers/psp.ts
```diff
@@ -200,90 +200,58 @@ const initializeEvmStates = async (state: State): Promise<State> => {
...
+  utils.logger.logPending(
+    failedCalls.flatMap((call) => call[0]),
+    baseLogOptions
+  );
 
-    if (isNil(data.apiValue)) {
-      utils.logger.warn('Failed to fetch API value. Skipping update...', templateLogOptions);
-      return acc;
-    }
+  const apiValuesBySubscriptionId = successfullCalls.reduce(
```
amarthadan, line 242:
```suggestion
  const apiValuesBySubscriptionId = successfulCalls.reduce(
```

src/handlers/psp.ts
```diff
@@ -200,90 +200,58 @@ const initializeEvmStates = async (state: State): Promise<State> => {
...
+  });
 
-    const templateLogOptions = buildLogOptions('additional', { templateId: data.templateId }, baseLogOptions);
+  const callApiResults = await Promise.all(apiValuePromises);
+  const successfullCalls = callApiResults.filter((call) => call[1].apiValue !== null);
+  const failedCalls = callApiResults.filter((call) => call[1].apiValue === null);
 
-    utils.logger.logPending(logs, templateLogOptions);
+  utils.logger.logPending(
+    successfullCalls.flatMap((call) => call[0]),
```
amarthadan, line 234:
```suggestion
    successfulCalls.flatMap((call) => call[0]),
```

yarn.lock
```diff
@@ -975,7 +973,7 @@
   resolved "https://registry.yarnpkg.com/@kwsites/promise-deferred/-/promise-deferred-1.1.1.tgz#8ace5259254426ccef57f3175bc64ed7095ed919"
   integrity sha512-GaHYm+c0O9MjZRu0ongGBRbinu8gVAMd2UZjji6jVmqKtZluZnptXGWhz1E8j8D2HJ3f/yMxKAUC0b+57wncIw==
 
-"@lifeomic/attempt@^3.0.0", "@lifeomic/attempt@^3.0.2":
+"@lifeomic/attempt@^3.0.0":
```
amarthadan, line 976: Why is it still here?

amarthadan (changes requested):

yarn.lock
Siegrift: Because we still have airnode packages in dependencies (despite using the postinstall trick).

package.json
```diff
@@ -32,6 +32,7 @@
     "sls:invoke-local:psp-beacon-update": "yarn build && sls invoke local --function psp-beacon-update --param='provider=local'",
     "setup:psp-local": "tsc && node scripts/psp-beacon-local-setup.js",
     "test": "jest",
+    "test:debug": "jest \"process-sponsor-wallet.test.ts\"",
```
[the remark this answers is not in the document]
amarthadan: I don't mind helper targets but the name is really confusing especially when you choose some random file and call it `debug` :smile: Either rename the target so it makes sense or remove it (anyone can just call jest directly if they want).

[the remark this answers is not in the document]
Siegrift: The point is not the name, but that this pattern is valid so that one doesn't have to google it everytime :D 
But I'll remove it.

vponline (approved):
LGTM  :+1:

amarthadan (approved):

PR #58, opened 2022-04-13
Add airnode-protocol-v1 dependency

vponline:
Use **airnode-protocol-v1** contracts to get ABI and bytecode


test/e2e/psp.feature.ts
```diff
@@ -40,19 +43,20 @@ const readBeaconValue = async (airnodeAddress: string, templateId: string, dapiS
...
   );
-  const dapiServerResponse = await dapiServer.connect(voidSigner).readWithDataPointId(beaconId);
 
-  return dapiServerResponse[0].toNumber();
+  try {
+    const dapiServerResponse = await dapiServer.connect(voidSigner).readDataFeedValueWithId(beaconId);
+    return dapiServerResponse.toNumber();
+  } catch (e) {
+    return null;
+  }
```
vponline, line 52: The function name has changed to `readDataFeedValueWithId` and it returns only the dapiServer value instead of an array: `[value, timestamp]`.

The new method also has added a check that the **dataFeed** exists instead of returning [0, timestamp] like the earlier version, hence the **try-catch**.

src/evm/initialize-provider.ts
```diff
@@ -13,11 +14,7 @@ export const initializeProvider = async (airnodeWalletMnemonic: string, provider
   const rrpBeaconServerAbi = new ethers.utils.Interface(protocol.RrpBeaconServerFactory.abi).format(
     ethers.utils.FormatTypes.minimal
   );
-
-  const dapiServerAbi = [
-    'function conditionPspBeaconUpdate(bytes32,bytes,bytes) view returns (bool)',
-    'function fulfillPspBeaconUpdate(bytes32,address,address,address,uint256,bytes,bytes)',
-  ];
+  const dapiServerAbi = new ethers.utils.Interface(DapiServerFactory.abi).format(ethers.utils.FormatTypes.minimal);
```
acenolaza, line 17: I wonder if we now need to keep formatting the abi as minimal 🤔 
I was only doing that because I wanted to keep the abi as short as possible when it was hardcoded.

test/e2e/psp.feature.ts
```diff
@@ -40,19 +43,20 @@ const readBeaconValue = async (airnodeAddress: string, templateId: string, dapiS
   const beaconId = hre.ethers.utils.keccak256(
     hre.ethers.utils.solidityPack(['address', 'bytes32'], [airnodeAddress, templateId])
   );
-  const dapiServerResponse = await dapiServer.connect(voidSigner).readWithDataPointId(beaconId);
 
-  return dapiServerResponse[0].toNumber();
+  try {
```
acenolaza, line 47: I'm a bit confused here. AFAIK the contract method can return an uint224 therefore it seems kind of risky that we do a `toNumber()` with a try-catch. Why do we even need to cast it to javascript number? Can we just keep using the ethers.BigNumber returned by the contract or maybe just cast it to string?

acenolaza (changes requested):
I'm really happy to see the hardcoded abi strings gone 🥳

test/e2e/psp.feature.ts
vponline: You make a good point, we can keep it as **BigNumber**

Siegrift (approved):
👍 LGTM, glad to +53, -3179 diff :D

acenolaza (approved):
LGTM 🏆

PR #59, opened 2022-04-14
Enable publishing to NPM and fix compilation location

aquarat:
This adds some changes to enable/improve NPM publishing.

Compilation output was being placed in `dist/src`, I've moved it to `dist`.
I've also changed the `package.json` index and types references so that consuming projects can re-use types by importing the package, rather than a path in the package.
Lastly I based the version off the existing version history, which was `0.0.1`, `0.0.4`, etc.

:)


Siegrift (approved):
LGTM

acenolaza (approved):
These changes are great. Thanks for taking care of this 🙏🏻 

One thing tho. I'm a bit worried that removing the `src/` segment from the path might break things like [this](https://github.com/api3dao/airkeeper/blob/npm-polishing-enablement/scripts/psp-beacon-local-setup.js#L6) and [this](https://github.com/api3dao/airkeeper/blob/npm-polishing-enablement/serverless.yml#L21-L37).

**Please** update those paths and run PSP locally (and ideally deploy it to aws) before merging.

amarthadan:
Also, make sure to try building and running the Docker container because I'm a bit worried this will break it.

aquarat:
Thanks for the comments all, I will double-check build and deployment before merging.

aquarat:
As per Slack, if I deploy using the latest `main` I get an error message:
```
Error:
CREATE_FAILED: IamRoleLambdaExecution (AWS::IAM::Role)
The policy failed legacy parsing (Service: AmazonIdentityManagement; Status Code: 400; Error Code: MalformedPolicyDocument; Request ID: x; Proxy: null)
```
If I set Resource in serverless.yml to '*' it deploys and the deployment runs correctly.

aquarat:
The `psp-beacon-update` function works, I'm waiting to see if `process-subscriptions` also works. I'll merge this when I have confirmation.

Separately it looks like I found some bugs along the way. I haven't addressed them here. I'll create a task for that.

aquarat:
[It's working nicely](https://monitor.api3.org/d-solo/SDapXdy7z/documentation-dashboard?orgId=1&var-chainId=3&var-beaconId=0xb5087bcd2a04f5653ad8a6b39fa4fb10fabbdec124b8b489b4d8d610e6392198&panelId=2), so I'll merge.

PR #60, opened 2022-04-18
Fix sls variables, dockerfile install script

vponline:
This fixes setting the serverless `stage` and `region`  variables with the CLI and fixes the dockerfile `yarn install` script to work with the airnode `postinstall` script.


aquarat (approved):
This works :rocket:, thanks for looking into this and also taking the time to fix the `README`.
Your attention to detail is appreciated :)

PR #76, opened 2022-08-29
Get timestamp from the chain instead of the system (PSP Airkeeper)

### LINKED ISSUE #68: Get timestamp from the chain instead of the system (PSP Airkeeper)
https://api3workspace.slack.com/archives/C02MJ5B03SN/p1653730318369929

acenolaza:
Closes #68


amarthadan (approved):
LGTM :+1:

PR #78, opened 2022-08-31
Better types for AWS handlers

### LINKED ISSUE #70: Better types for AWS handlers PSP Airkeeper
You can take a look at airnode-deployer to see how to type them.

acenolaza:
Closes #70

### DIFF
--- a/src/handlers/process-subscriptions.ts
+++ b/src/handlers/process-subscriptions.ts
@@ -2,10 +2,10 @@ import * as utils from '@api3/airnode-utilities';
 import { goSync } from '@api3/promise-utils';
 import isNil from 'lodash/isNil';
 import { loadAirnodeConfig } from '../config';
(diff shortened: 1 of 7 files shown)


src/handlers/process-subscriptions.ts
```diff
@@ -59,13 +59,14 @@ export const processSubscriptions = async (
   });
 };
 
-export const handler = async ({
-  providerSponsorSubscriptions,
-  baseLogOptions,
-}: {
+export const handler = async (payload: {
```
acenolaza, line 62: I played around with the idea of adding the aws context object to the `state` var in psp.ts and then forward it here after doing `lambda.invoke()` but I dropped the idea after seeing that it didn't add much value.

amarthadan (approved):
LGTM

PR #79, opened 2022-09-13
73 protocol constants

### LINKED ISSUE #73: Use protocol constants from Airnode inside Airkeeper
We have added protocol indices as part of v0.8 Airnode release. We should use these constants inside Airkeeper.

See: https://github.com/api3dao/airnode/issues/1195

acenolaza:
Closes #73 
Similar to [236](https://github.com/api3dao/airseeker/pull/236) in airseeker

### DIFF
--- a/scripts/psp-beacon-local-setup.js
+++ b/scripts/psp-beacon-local-setup.js
@@ -3,12 +3,12 @@ const path = require('path');
 const ethers = require('ethers');
 const abi = require('@api3/airnode-abi');
 const { evm } = require('@api3/airnode-node');
+const { PROTOCOL_IDS } = require('@api3/airnode-protocol');
 const {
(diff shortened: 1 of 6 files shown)


amarthadan (approved):
I presume this is correct as it was `2` before but out of curiosity, what is the `PROTOCOL_IDS.AIRKEEPER` for? And Airkeeper RRP?

acenolaza:
> I presume this is correct as it was `2` before but out of curiosity, what is the `PROTOCOL_IDS.AIRKEEPER` for? And Airkeeper RRP?

For sponsor wallets using RrpBeaconServer to update beacons we used a throwaway protocol_id of [`12345`](https://github.com/api3dao/airkeeper/blob/71-start-time-log-psp/src/handlers/rrp.ts#L187). I believe the original idea was to switch to DapiServer once it was ready.

PR #80, opened 2022-09-13
Moves script files into the /scripts directory

### LINKED ISSUE #72: Move scripts to `/scripts` directory | PSP Airkeeper
There are files load-secrets-env.js and export-aws-env.sh that should be moved to the /scripts directory.

As part of this move, both should be rewritten in TypeScript.

acenolaza:
Closes #72


amarthadan (changes requested):
I think you also need to modify the content of `export-aws-env.sh` (can't comment there as there were no changes)
```
test -f config/aws.env && export $(egrep -v '^#' config/aws.env | xargs)
vs
test -f ../config/aws.env && export $(egrep -v '^#' ../config/aws.env | xargs)
```

acenolaza:
> I think you also need to modify the content of `export-aws-env.sh` (can't comment there as there were no changes)
> 
> ```
> test -f config/aws.env && export $(egrep -v '^#' config/aws.env | xargs)
> vs
> test -f ../config/aws.env && export $(egrep -v '^#' ../config/aws.env | xargs)
> ```

Good catch! 🔥

acenolaza:
@amarthadan long story short, I ended up taking the Airseeker approach of just asking the user to set the AWS creds env vars before running the `sls deploy` command in the terminal.

Running the export-aws-env.sh script from a yarn script defined in package.json turned out to be a nightmare and ended up realizing that there might be a reason for this script to be gone from all other repos and this is that it only works on linux on a specific shell. Then I also removed the STAGE and REGION env vars to again, be consistent with Airseeker.

This should now be good to go.

README.md
```diff
@@ -84,7 +83,7 @@ docker run -it --rm ^
 api3/airkeeper:latest deploy --stage dev --region us-east-1
 ```
 
-### Removal
+### Deploy
```
amarthadan, line 86:
```suggestion
### Remove
```

README.md
```diff
@@ -104,6 +103,16 @@ docker run -it --rm ^
 api3/airkeeper:latest remove --stage dev --region us-east-1
 ```
 
+## Deploy from CLI
+
+Airkeeper can also be deployed and removed form AWS by just running the following commands in the terminal:
```
amarthadan, line 108:
```suggestion
Airkeeper can also be deployed and removed from AWS by just running the following commands in the terminal:
```

README.md
```diff
@@ -128,10 +137,11 @@ Follow these steps to run Airkeeper locally:
 ### Running Airkeeper on AWS Lambda
 
 Airkeeper is meant to be deployed to the AWS lambda service and for this you will need to add your credentials to the
-`config/aws.env` file. Once done, the `export-aws-env.sh` script can be used to load them into the environment.
+`config/aws.env` file. Once done, the [`export-aws-env.sh`](scripts/export-aws-env.sh) script can be used to load them
+into the environment.
```
amarthadan, line 141: I'm a bit confused. After reading your comment I thought that you've removed the script altogether.

package.json
```diff
@@ -35,18 +36,18 @@
...
     "prettier:check": "prettier --check \"./**/*.{js,sol,ts,md,yml,json}\"",
     "prettier:write": "prettier --write \"./**/*.{js,sol,ts,md,yml,json}\" --ignore-unknown",
-    "sls:config": "sls config credentials --provider aws --key $AWS_ACCESS_KEY_ID --secret $AWS_SECRET_ACCESS_KEY",
-    "sls:deploy": "yarn build && sls deploy --region $REGION --stage $STAGE",
-    "sls:remove": "sls remove --region $REGION --stage $STAGE",
+    "sls:deploy": "sls deploy",
+    "sls:remove": "sls remove",
     "sls:invoke:rrp-beacon-update": "sls invoke --function rrp-beacon-update",
-    "sls:invoke-local:rrp-beacon-update": "yarn build && sls invoke local --function rrp-beacon-update",
+    "sls:invoke-local:rrp-beacon-update": "sls invoke local --function rrp-beacon-update",
```
amarthadan, line 47: Why did you remove the `build` part? Don't you need to build the lambda before invoking it?

README.md
acenolaza: Good catch. I forgot to remove this part after removing the script 👏🏻

amarthadan: But the script itself is still there. At least according to GitHub :smile:

package.json
acenolaza: yes, also good catch! I was playing around with `pre` and `post` scripts. I added a `predeploy:sls` script where I was planning to do the `yarn build` but then I didn't like the `pre` script name being so long for running the functions locally so I dropped the idea but forgot to add the `yarn build &&` back.

README.md
acenolaza: LOL 🤦🏻 ... deleting file

amarthadan (approved):
:+1:

PR #82, opened 2022-09-14
Generate airkeeper config using beacon data

Ashar2shahid:
Opening a draft PR, The goal is to have a single beacon data file that doesn't need scripts to fill (i.e generate templateIds,beaconIds,encode parameters etc) to generate the final `airkeeper.json`.  This is based on the #62 branch.


scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+    const options: ChainOptions = {
+      fulfillmentGasLimit: 500000,
+      gasPriceOracle: [
+        {
+          gasPriceStrategy: 'constantGasPrice',
+          gasPrice: {
+            value: 10,
+            unit: 'gwei',
+          },
+        } as const,
```
acenolaza, line 42: @vponline suggested that I add more strategies to the example config file. Maybe that applies here as well, right? 

These might need to go above the `constantGasPrice`:

```
          {
            "gasPriceStrategy": "latestBlockPercentileGasPrice",
            "percentile": 60,
            "minTransactionCount": 20,
            "pastToCompareInBlocks": 20,
            "maxDeviationMultiplier": 2
          },
          {
            "gasPriceStrategy": "providerRecommendedGasPrice",
            "recommendedGasPriceMultiplier": 1.2
          },
```

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+      [`provider_${sanitiseFilename(chainName).replace(/\-/g, '_')}`]: {
+        url: `\${${sanitiseFilename(chainName).replace(/\-/g, '_')}_PROVIDER_URL}`.toUpperCase(),
+      },
+    };
+
+    return {
+      contracts,
+      id: `${chainId}`,
+      providers,
+      type: 'evm' as const,
```
acenolaza, line 55: Do you mind explaining why we need this casts `as const` here and there?

acenolaza: The whole return can even be a `ChainConfig` object type

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+        const endpointId = ethers.utils.keccak256(
+          ethers.utils.defaultAbiCoder.encode(
+            ['string', 'string'],
+            [beacon.template.oisTitle, beacon.template.endpointName]
+          )
+        );
+        const templateId = ethers.utils.solidityKeccak256(['bytes32', 'bytes'], [endpointId, templateParameters]);
+        const threshold = ethers.BigNumber.from(100000000)
+          .mul(chain.updateConditionPercentage * 100)
+          .div(10000);
```
acenolaza, line 113: Do we really need to multiply by 100 and then divide by 10000? Can't we just divide by 100?

```suggestion
        const threshold = ethers.BigNumber.from(100_000_000)
          .mul(chain.updateConditionPercentage)
          .div(100);
```

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+            parameters,
+            airnodeAddress,
+            templateId,
+            conditions: encodedBeaconUpdateSubscriptionConditions,
+            relayer: airnodeAddress,
+            sponsor,
+            requester: dapiServerAddress,
+            fulfillFunctionId: dapiServerInteface.getSighash('fulfillPspBeaconUpdate'),
+          },
+        };
```
acenolaza, line 159: This could also be a [Subscription](https://github.com/api3dao/airkeeper/blob/generate-airkeeper-config/src/validator/validator.ts#L99) object type

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+      })
+    )
+    .reduce((subscriptionsObject, subscription) => ({ ...subscriptionsObject, ...subscription }), {});
+
+  const airkeeperTriggers: Triggers = {
+    rrpBeaconServerKeeperJobs: [],
```
acenolaza, line 165: Should we add a TODO here for these triggers?

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+      const templateId = ethers.utils.solidityKeccak256(['bytes32', 'bytes'], [endpointId, encodedParameters]);
+
+      return {
+        ...templateObj,
+        [templateId]: {
+          endpointId,
+          encodedParameters,
+        },
+      };
+    }, {}) as Templates;
```
acenolaza, line 185: Now that I think about it we could add the type to the `airkeeperTemplates` instead of explicit casting, right?
```suggestion
  const airkeeperTemplates: Templates = Object.values(ConfigurationData.beacons)
    .map((beacon: any) => beacon.template)
    .reduce((templateObj: any, template: any) => {
      const endpointId = ethers.utils.keccak256(
        ethers.utils.defaultAbiCoder.encode(['string', 'string'], [template.oisTitle, template.endpointName])
      );
      const encodedParameters = encode(template.decodedParameters);
      const templateId = ethers.utils.solidityKeccak256(['bytes32', 'bytes'], [endpointId, encodedParameters]);

      return {
        ...templateObj,
        [templateId]: {
          endpointId,
          encodedParameters,
        },
      };
    }, {});
```

acenolaza: Same for all other objects

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+    endpoints: AirkeeperEndpoints,
+    ois: Object.values(ConfigurationData.ois),
+    apiCredentials: apiCredentials,
+  };
+
+  writeJsonFile(join(__dirname, 'airkeeper.json'), airkeeper);
+  writeJsonFile(join(__dirname, 'secrets'), { filename: '.env', content: airkeeperSecretsArray.join('\n') });
+  writeJsonFile(join(__dirname, 'aws'), { filename: '.env', content: awsSecretsArray.join('\n') });
+
+  console.log(airkeeper);
```
acenolaza, line 216: I'm guessing this is just for testing right?

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+    subscriptions: airkeeperSubscriptions,
+    templatesV1: airkeeperTemplates,
+    endpoints: AirkeeperEndpoints,
+    ois: Object.values(ConfigurationData.ois),
+    apiCredentials: apiCredentials,
+  };
+
+  writeJsonFile(join(__dirname, 'airkeeper.json'), airkeeper);
+  writeJsonFile(join(__dirname, 'secrets'), { filename: '.env', content: airkeeperSecretsArray.join('\n') });
+  writeJsonFile(join(__dirname, 'aws'), { filename: '.env', content: awsSecretsArray.join('\n') });
```
acenolaza, line 214: Should these be created under `../../config` instead?

scripts/config/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+      files: '*.md',
+      options: {
+        parser: 'markdown',
+        proseWrap: 'always',
+      },
+    },
+  ],
+} as any;
+
+export const PROTOCOL_ID_PSP = '2';
```
acenolaza, line 25: This should be imported from `airnode-protocol`

acenolaza (changes requested):
Looking good but added a few comments.

Also, should we prompt the user to select a specific api provider for generating triggers? Should we ask the user if he wants to create RRP or PSP triggers? Both maybe?

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
+import { join } from 'path';
+import { ethers } from 'ethers';
+import { encode } from '@api3/airnode-abi';
+import { AirnodeRrpAddresses } from '@api3/airnode-protocol';
+import { readOperationsRepository } from '@api3/operations/dist/utils/read-operations';
```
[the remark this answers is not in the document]
Ashar2shahid: yeah, can update the imports after we do

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+  runAndHandleErrors,
+  writeJsonFile,
+  getDapiServerInterface,
+  sanitiseFilename,
+} from './utils';
+import { Config, ChainConfig, NodeSettings, Subscriptions, Triggers, Templates, Endpoints } from '../../src';
+
+const main = async () => {
+  const operationsRepository = readOperationsRepository();
+  const ConfigurationData = readConfigurationData();
```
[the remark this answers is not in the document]
Ashar2shahid: so I was thinking that this can be used to create operation configs and also local configs. I made a new update where its an option todo both

scripts/config/create-config.ts
```diff
@@ -0,0 +1,221 @@
...
+  const apiChains = [
+    ...new Set(Object.values(ConfigurationData.beacons).flatMap((beacon: any) => Object.keys(beacon.chains))),
+  ];
+
+  const chains: ChainConfig[] = apiChains.map((chainName) => {
+    const chainId = parseInt(operationsRepository.chains[chainName].id);
+    const contracts = {
+      AirnodeRrp: AirnodeRrpAddresses[chainId] || '',
+      RrpBeaconServer: operationsRepository.chains[chainName].contracts.RrpBeaconServer || '',
+      DapiServer: operationsRepository.chains[chainName].contracts.DapiServer || '',
```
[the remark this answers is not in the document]
Ashar2shahid: its reading them from there atm but yeah we need to create a task to add them to operations

scripts/config/create-config.ts
Ashar2shahid: done

scripts/config/create-config.ts
Ashar2shahid: yeah, I realized it doesn't need to be const. Fixed

scripts/config/create-config.ts
Ashar2shahid: yeah because if the value is something like 0.01 then it doesn't work as expected

scripts/config/create-config.ts
Ashar2shahid: subscription object is only this part:
```
{
            chainId: `${chainId}`,
            parameters,
            airnodeAddress,
            templateId,
            conditions: encodedBeaconUpdateSubscriptionConditions,
            relayer: airnodeAddress,
            sponsor,
            requester: dapiServerAddress,
            fulfillFunctionId: dapiServerInteface.getSighash('fulfillPspBeaconUpdate'),
}
```
so we are returning a `Subscriptions` object

scripts/config/create-config.ts
Ashar2shahid: yeah good idea

scripts/config/create-config.ts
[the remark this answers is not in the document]
acenolaza: just to confirm, are we then OK with `RrpBeaconServer` always being `''`?

scripts/config/create-config.ts
```diff
@@ -0,0 +1,263 @@
...
+    ...oisSecrets,
+  ];
+
+  const awsSecretsArray = [
+    `AWS_ACCESS_KEY_ID=`,
+    `AWS_SECRET_ACCESS_KEY=`,
+    `AWS_SESSION_TOKEN=`,
+    `REGION=us-east-1`,
+    `STAGE=dev`,
+  ];
```
acenolaza, line 143: I've removed these 2 [here](https://github.com/api3dao/airkeeper/commit/34c83637f1d1b985ed39da5fc44b60ce22f40a92)

scripts/config/create-config.ts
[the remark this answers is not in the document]
acenolaza: we also referencing DapiServer address again in line 174

acenolaza (approved):
LGTM

PLEASE test deploying to AWS to confirm that nothing has broken

scripts/config/create-config.ts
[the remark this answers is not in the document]
acenolaza: so what would be the use case for local data? just setting up an airkeeper.json for local testing?

acenolaza:
I forgot to mention that this needs to be documented and added to the README

scripts/config/create-config.ts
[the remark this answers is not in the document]
Ashar2shahid: yeah, setting it up for local testing

